PR #45, opened 2021-04-21
Pip build

Spartee:
This largely addresses the automation of the entire pip installation process. The build time is much shorter as we got rid of some unnecessary steps. 

Using this, SmartRedis can be completely installed into a python env through ``pip install .``. 

Notably the python build also triggers the build of the static library so that is included as well. 

I am also going add some github workflows in this PR to run manylinux builds that we can host on PyPI so users can install SmartRedis with ``pip install SmartRedis``. In the case where we don't have a built wheel for the user arch, the tarball will be downloaded and built on the user system using ``setup.py``.


al-rigazzi (approved):
LGTM

PR #53, opened 2021-05-17
Fix error in Python exists functions

mellis13:
Currently the ``tensor_exists``, ``model_exists``, and ``key_exists`` return ``None`` instead of a boolean value.  This PR fixes this behavior.


al-rigazzi (commented):
LGTM, good catch!

al-rigazzi (approved):

Spartee (approved):
LGTM

Issue #64, opened 2021-06-09
Remove output related to TimeoutError and IOError


mellis13:
# Description
Currently warnings are printed if an IOError or TimeoutError are encountered when executing a command.  This can lead to too much output when many clients are active and it is hard to read.  We will only have output on failed execution after all attempts are made.

# Justification
This will improve user experience.

# Implementation Strategy
We will remove the warning output.

PR #65, opened 2021-06-09
Remove warning output for TimeoutError and IOError

mellis13:
This PR removes the output warning associated with failed command execution.  Only the command will be printed after all attempts have failed.  Additionally, the delay in successive attempts is made consistent between TimeoutError and IOError.  This address #64 .


al-rigazzi (approved):
LGTM

PR #67, opened 2021-06-16
Fixes build of GPU dependencies for RedisAI

al-rigazzi:
This is a quick fix for building GPU dependencies for RedisAI. Currently, the script builds CPU dependencies, and the build of RedisAI for GPU fails.


mellis13 (approved):
Great catch!  Thanks for fixing.  Looks good to merge.

PR #70, opened 2021-06-18
Improve memory management for get_tensor()

mellis13:
In this PR, the memory management for ``Client.get_tensor()`` and ``DataSet.get_tensor()`` have been modified and improved.  The following changes to memory management have been made:

- ``Client.get_tensor()`` calls from the Python client will now pass memory management to the numpy array through a PyBind ``capsule``.   This means that the memory created during a ``Client.get_tensor()`` call from Python will be freed when the numpy array is deleted in Python (either manually or automatically based on scope).  This allows for repeated calls of ``Client.get_tensor()`` from Python without worrying about massive memory consumption.
- ``DataSet.get_tensor()`` will have the same memory management behavior as ``Client.get_tensor()`` when called from Python (i.e. managed by the numpy array).
- ``DataSet.get_tensor()`` now returns a deep copy of the tensor data to all compiled clients (C,C++,Fortran).  In the past, only Fortran received a deep copy.  In the compiled clients, memory is not freed from a ``get_tensor()`` call until the ``Client`` or ``DataSet`` goes out of scope or is deleted.

Additional small changes were made in this PR as small issues were discovered:

- The Fortran test suite now defaults to ``.False.`` for using clusters if the testing environment variable is not present.  Previously the behavior was undefined.
- PYTHONFAULTHANDLER=1 has been added to python testing calls to provide better failure inspection
- ``DataSet`` will throw a ``std::runtime_error`` if an attempt is made to send an empty DataSet
- A bug in ``DataSet`` was fixed that prevented from sending and receiving ``DataSet`` without tensors


src/cpp/client.cpp
```diff
@@ -932,4 +884,78 @@ void Client::_unpack_dataset_metadata(DataSet& dataset,
...
+
+    if(dims.size()<=0)
+        throw std::runtime_error("The number of dimensions of the "\
+                                "fetched tensor are invalid: " +
+                                std::to_string(dims.size()));
+
+    for(size_t i=0; i<dims.size(); i++) {
+        if(dims[i]<=0) {
+        throw std::runtime_error("Dimension " +
+                                 std::to_string(dims[i]) +
```
al-rigazzi, line 912: Shouldn't this be just `i` instead of `dims[i]`?

src/cpp/client.cpp
```diff
@@ -932,4 +884,78 @@ void Client::_unpack_dataset_metadata(DataSet& dataset,
...
+                                       type, MemoryLayout::contiguous);
+            break;
+        case TensorType::uint8 :
+            ptr = new Tensor<uint8_t>(g_key, (void*)blob.data(), dims,
+                                      type, MemoryLayout::contiguous);
+            break;
+        default :
+            throw std::runtime_error("The tensor could not be "\
+                                     "retrieved in "\
+                                     "client.get_tensorbase_obj().");
```
al-rigazzi, line 957: Can we make this error message more informative? Something about the fact the type was was not recognized as a TensorType or similar?

al-rigazzi (commented):
Looks good, just have a couple of comments.

src/cpp/client.cpp
mellis13: Yes, you are correct.  The error output should be ``i``.  Thanks!

PR #72, opened 2021-06-29
Fix Tensor and TensorBase move assignment operator

EricGustin:
Fix for bugs relating to the `Tensor` and `TensorBase` move assignment operators and move constructors. `Tensor`'s move assignment operator and move constructor are now working as intended, and `_data` for rvalue reference is being freed.


mellis13 (approved):
Looks good!  Thank you for catching this.

PR #73, opened 2021-07-01
Port to NCAR cheyenne

jedwards4b:
This is what I needed to get a successful build on the NCAR cheyenne system.


jedwards4b:
slurm srun is in the path on cheyenne but cannot be used, so you need to remove it from your path
before you can run test-fortran.

Spartee:
Hi @jedwards4b, for the python portion, would it instead be possible to activate a python environment where ``python3`` is automatically symlinked to ``python``. Doing so would emulate how we expect users to utilize the makefile.

jedwards4b:
@Spartee Tried a second time, using a python environment did solve the python3 issue, the 
only change I ended up needing was in build_deps.sh  The variable LIBRARY_PATH is set by cheyenne modules, explicitly setting it here.

jedwards4b:
But after fixing a merge conflict there are no changes.  closing this PR

PR #75, opened 2021-07-02
Revert "Improve memory management for get_tensor()"

mellis13:
Unexpected failure in github action.


mellis13:
The failing github test is not due to this PR.  It is because of a failure to pin to a version of PyBind. A separate PR to pin to a version of PyBind has been submitted.

PR #76, opened 2021-07-02
Pin PyBind to v2.6.2.

mellis13:
Currently, we are not pinned to a version of PyBind, and because of this, the github automated tests started failing.  This PR pins SmartRedis to use PyBind v2.6.2.


al-rigazzi (approved):
LGTM

PR #78, opened 2021-07-05
hiredis path fix

al-rigazzi:
On Summit, the env variable `LIBRARY_PATH` is defined at system level (similarly to `LD_LIBRARY_PATH`). This variable clashes with one variable in the hiredis build `Makefile`. The result is that libraries end up in the wrong path and `build_deps.sh` cannot find them.


mellis13 (approved):
Thanks for investigating and fixing this.

PR #80, opened 2021-07-07
Add CommandList and Command copy and move operators

### LINKED ISSUE #60: CommandList and Command copy and move operators
# Description
``CommandList`` is used to group multiple commands to be executed by ``RedisCluster`` or ``Redis``.  The ``Command`` objects contained in the ``CommandList`` are dynamically allocated and currently only pointers to the dynamically created ``Command`` objects are stored.  The copy and move operators for the ``CommandlList`` are not implemented.  These operators should be defined and implemented.

# Justification
As issue #43 is addressed, the ``CommandList`` object should be completed.  The default operators defined by the compiler may not produce the intended behavior (i.e. copy operations would be a shallow copy of the underlying ``Command``).   Default move operators should be sufficient, but should be specified as such in the header with ``default``.

# Implementation Strategy

- [x] Implement copy operators for `Command`
- [x] Implement copy operators for `CommandList`
- [x] Define `CommandList` move operators as default
- [x] Define and document behavior of copy operators

EricGustin:
This PR addresses and closes #60 by implementing copy operators for `Command` and `CommandList`, and defining `CommandList` move operators as default.


src/cpp/command.cpp
```diff
@@ -30,6 +30,45 @@
...
 using namespace SmartRedis;
 
+Command::Command(const Command& cmd)
+{
+    this->_fields = std::vector<std::string_view>(cmd._fields);
+    this->_cmd_keys = std::unordered_map<std::string_view, size_t>(cmd._cmd_keys);
+    std::vector<char*>::const_iterator local_fields_it = cmd._local_fields.cbegin();
+    std::vector<char*>::const_iterator local_fields_it_end = cmd._local_fields.cend();
+    while(local_fields_it != local_fields_it_end) {
+        int field_size = std::strlen(*local_fields_it);
```
mellis13, line 40: Relying on ``strlen`` is problematic because fields are not guaranteed to be null-terminated or might contain null-terminating characters in the middle of the field (i.e. tensor data).  We might need to update/rearrange the Command data structures so we more easily track the length of the allocated memory along with the pointer.

EricGustin: So then we need to keep track of the length of the allocated memory by storing the length whenever a field is pushed to the back of `_local_fields`. What do you think about adding a new data member that is a vector of ints to the `Command` data structure? The ith element in this vector corresponds to the length of the allocated memory of the ith element in `_local_fields`. @mellis13

mellis13 (approved):
Looks good!  Go ahead and merge when ready.

Issue #71, opened 2021-06-21
Implement functions for commands without keys


mellis13:
# Description
The ``Command`` object and ``RedisServer`` object  were originally written for keyed commands.  However, as the ability to execute commands like ``INFO`` is added to the Python client (see #1 ), these objects need to handle non-keyed commands that might address a single database node or the entire cluster.

# Justification
This is a prerequisite feature in order to be able to accomplish #1 which will give users better view of database characteristics and performance.

# Implementation Strategy
The design changes to ``Command`` and ``RedisServer`` should allow for the following functionality:
- A non-keyed command working for a Redis non-cluster database
- A non-keyed command addressable to a single node in a cluster database
- A non-keyed command addressable to all nodes in a cluster database (i.e. collect replies from all nodes in the database)

PR #83, opened 2021-07-12
Implement commands without keys; Add INFO command to Python Client

ctandon11:
The `Command` and `RedisServer` ocan now handle non-keyed commands 
that might address a single database node or the entire cluster. This involved
adding an `is_addressable()` function to check if database address is valid.

This will be used as we add the ability to execute commands in the Python Client.

The Redis `INFO` command is now executable in the Python Client. It returns 
information and statistics about given database node(s).

This PR addresses issue #71 as well as a portion of issue #1 .


include/client.h
```diff
@@ -549,6 +552,21 @@ class Client
...
+        */
+        parsed_map get_db_node_info(std::string address);
+
+        /*!
+        *   \brief Parse database node information from get_db_node_info()
+        *          into a nested unordered_map
+        *   \param info database node information
+        *   \returns parsed_map containing the database node information
+        */
+        parsed_map parse_db_node_info(std::string info);
```
mellis13, line 568: I wonder if this should be made into a private method ``_parse_db_node_info`` since it doesn't seem like a user would be calling it and it is currently called inside of ``get_db_node_info()``.

include/command.h
```diff
@@ -200,6 +200,29 @@ class Command
...
+        *          to be executed on
+        *   \param address Address of database
+        *   \param port Port of database
+        */
+        void set_exec_address_port(std::string address,
+                                   uint64_t port);
+
+        /*!
+        *   \brief Get address that command will be
+        *          to be executed on
```
mellis13, line 214: Maybe we can indicate what would be returned if the ``set_exec_address_port()`` method had never been called?  If we go with the default value of a ``std::string``, maybe we should say an empty string is returned.

include/command.h
```diff
@@ -200,6 +200,29 @@ class Command
...
+        *   \return std::string of address
+        */
+        std::string get_address();
+
+        /*!
+        *   \brief Get port that command will be
+        *          to be executed on
```
mellis13, line 221: I think we should also specify the behavior of what gets returned in the case an address and port hasn't been set on the ``Command``.

include/command.h
```diff
@@ -269,6 +292,16 @@ class Command
...
 
+        /*!
+        *   \brief Address of database node
+        */
+        std::string _address;
+
+        /*!
+        *   \brief Port of database node
+        */
+        uint64_t _port;
```
mellis13, line 303: I think we will want to specify a default value of ``_port`` that is something that can be distinguished as ``Command`` does not have an address and port for execution.  Reading more about standards, it seems like ports are constrained to values that fit into a 16-bit unsigned integer (maybe we change the data type throughout the code to reflect this?).

include/redisserver.h
```diff
@@ -275,6 +284,11 @@ class RedisServer {
         */
         std::string _get_ssdb();
 
+        /*!
+        *   \brief Unordered map of address:port to DBNode in the cluster
+        */
+        std::unordered_map<std::string, DBNode*> _valid_nodes;
```
mellis13, line 290: Maybe a better variable name would be ``_address_node_map``?

include/redisserver.h
```diff
@@ -99,6 +100,14 @@ class RedisServer {
          */
         virtual bool model_key_exists(const std::string& key) = 0;
 
+        /*!
+         *  \brief Check if address is valid
```
mellis13, line 104: Maybe clarify 'valid' -- ``Check if address and port maps to database node``

src/cpp/client.cpp
```diff
@@ -665,6 +665,62 @@ void Client::use_tensor_ensemble_prefix(bool use_prefix)
     this->_use_tensor_prefix = use_prefix;
 }
 
+parsed_map Client::get_db_node_info(std::string address)
+{
+    std::string host = address.substr(0, address.find(":"));
+    uint64_t port = std::stoul (address.substr(address.find(":") + 1),
+                                nullptr, 0);
```
mellis13, line 672: Given that this is a user-facing function, we should do some checking on host and port to make sure we are getting a reasonable value, and if not, issue a helpful error.

src/cpp/client.cpp
```diff
@@ -665,6 +665,62 @@ void Client::use_tensor_ensemble_prefix(bool use_prefix)
...
+    Command cmd;
+    cmd.set_exec_address_port(host, port);
+    cmd.add_field("INFO");
```
mellis13, line 675: We should add another field "everything" so that we get all categories are returned (i.e. have the Redis command be ``INFO everything``.

src/cpp/client.cpp
```diff
@@ -665,6 +665,62 @@ void Client::use_tensor_ensemble_prefix(bool use_prefix)
...
+    CommandReply reply = this->_run(cmd);
+    // Put in a check here to make sure it is a str field or
+	// or throw error? this is done in reply.str()
```
mellis13, line 678: We can take this comment out.  ``CommandReply::str()`` will throw an error if the reply is not a string type.

src/cpp/client.cpp
```diff
@@ -665,6 +665,62 @@ void Client::use_tensor_ensemble_prefix(bool use_prefix)
...
+    parsed_map info_map = this ->
+        parse_db_node_info(std::string(reply.str(), reply.str_len()));
+    return info_map;
```
mellis13, line 681: We don't to save an intermediate variable ``info_map`` here.  We could just return the function return value.

src/cpp/client.cpp
```diff
@@ -665,6 +665,62 @@ void Client::use_tensor_ensemble_prefix(bool use_prefix)
...
+}
+
+parsed_map Client::parse_db_node_info(std::string info)
+{
+    parsed_map info_map;
+
+    std::string delim = "\r\n";
+    std::string currKey = "";
+    auto start = 0U;
+    auto end = info.find(delim);
```
mellis13, line 691: I think we can just use ``size_t`` here instead of ``auto``

src/cpp/client.cpp
```diff
@@ -665,6 +665,62 @@ void Client::use_tensor_ensemble_prefix(bool use_prefix)
...
+    return info_map;
+}
+
+parsed_map Client::parse_db_node_info(std::string info)
+{
+    parsed_map info_map;
+
+    std::string delim = "\r\n";
+    std::string currKey = "";
+    auto start = 0U;
```
mellis13, line 690: I think we can just use ``size_t`` here instead of ``auto``

src/cpp/client.cpp
```diff
@@ -665,6 +665,62 @@ void Client::use_tensor_ensemble_prefix(bool use_prefix)
...
+        end = info.find(delim, start);
+        if (line.length() == 0)
+            continue;
+        if (line[0] == '#')
+            currKey = line.substr(2);
+        else
+        {
+            std::size_t separatorIdx = line.find(':');
+            if (info_map.find(currKey)==info_map.end())
+                info_map[currKey] = {};
```
mellis13, line 706: Should this ``if`` statement be nested under the ``if (line[0] == '#')`` to avoid doing this check for every entry in a section?

src/cpp/client.cpp
```diff
@@ -665,6 +665,62 @@ void Client::use_tensor_ensemble_prefix(bool use_prefix)
...
+            info_map[currKey][line.substr(0, separatorIdx)] =
+                                line.substr(separatorIdx+1);
+        }
+    }
+    std::string line = info.substr(start);
+    if (line.length() == 0)
```
mellis13, line 712: Instead of an alternate return path, can we do ``if (line.length()>0)`` and include the last entry in that if statement?  That way we don't have alternate return paths.

src/cpp/rediscluster.cpp
```diff
@@ -53,7 +53,16 @@ RedisCluster::~RedisCluster()
 
 CommandReply RedisCluster::run(Command& cmd)
 {
-    std::string db_prefix = this->_get_db_node_prefix(cmd);
+    std::string db_prefix;
+
+    if (cmd.has_keys())
+        db_prefix = this->_get_db_node_prefix(cmd);
+    else if (this->is_addressable(cmd.get_address(), cmd.get_port()))
```
mellis13, line 60: I wonder if we should put the address, port case first so that if an address and port are specified, it always overrides the presence of keys.

tests/python/test_address.py
```diff
@@ -13,3 +13,21 @@ def test_address(use_cluster):
 
     # check if SSDB was set anyway
     assert os.environ["SSDB"] == ssdb
+
+
+def test_dbnode_info_command(use_cluster):
```
mellis13, line 18: We should probably start a new test file for all of the non-keyed commands.

include/client.h
ctandon11: yes I agree I'll switch it to private.

include/command.h
ctandon11: yep ill add that an empty string is returned if address isn't set

include/command.h
ctandon11: Ok i will change the `_port` to be a `uint16_t`. The default for an `unit16_t` is `0`. Do you think there would be any case where the `_port` would be set to `0`? If not, we can keep that as the default.

include/commandreplyparser.h
```diff
@@ -34,6 +34,9 @@
 #include "tensorbase.h"
 #include "enums/cpp_tensor_type.h"
 
+using inner_map = std::unordered_map<std::string, std::string>;
+using parsed_map = std::unordered_map<std::string, inner_map>;
```
mellis13, line 38: Maybe ``parsed_reply_map`` is a better type name?

mellis13 (approved):
Great PR!  Just rebase with ``develop`` and make sure tests pass.

Issue #82, opened 2021-07-08
DataSet rename() and delete() CROSSSLOT error with ensembles


mellis13:
# Description
Currently the ``DataSet`` ack key is the set to the name of the ``DataSet``.  In the case where the client is used in conjunction with SmartSim ensemble capabilities (e.g. ``SSKEYIN`` and/or ``SSKEYOUT`` are set), there will be an error because the ack key will take the form ``ensemble_name.dataset_name`` while all other keys are of the form ``{dataset_name}.meta`` or ``{dataset_name}.tensor_name``.  On clusters, this will cause a CROSSSLOT error because the ack key will not be in the same hash slot as the other fields.

A short-term fix for this problem could be just to separate the manipulations of the ack key for rename and delete such that there is no CROSSLOT error.  This works for the current implementation because ``DataSet`` put and get commands internally are not pipelined, so we do not need to ensure everything goes to the same hash slot or database node.  However, in the future we should have all ``DataSet`` entities go to the same hash slot.  Possible future (more robust) solutions include:

- Put the ack key/value into the redis hashes object that stores all of the ``DataSet`` metadata.
- Get rid of the ack key and use the redis hash object key as the ack key.

If this either of the more robust fixes is implemented, the process for checking the existence of a DataSet will need to be changed.

# How to reproduce
Export the following variables before running the ``client_test_delete_dataset`` test:

```bash
export SSKEYIN="test_ensemble"
export SSKEYOUT="test_ensemble"
```

This will produce ``terminating with uncaught exception of type sw::redis::ReplyError: CROSSSLOT Keys in request don't hash to the same slot``.

# Expected behavior
The ack key should work correctly with and without the ensemble functionality activated.

# System
All systems

PR #84, opened 2021-07-13
Fix CROSSSLOT error for DataSet deletion with ensembles

mellis13:
This PR fixes the bug reported in #82 .  The deletion of the DataSet ack key is split into a separate command to avoid CROSSSLOT errors when ensemble prefixing is active.  A test has been added to verify that the bug is fixed.


al-rigazzi (approved):
LGTM!

PR #93, opened 2021-07-19
add an is_initialized method to the fortran client

jedwards4b:
Add a method to fortran client to check if client%initialize has already been called.

### DIFF
--- a/src/fortran/client.F90
+++ b/src/fortran/client.F90
@@ -51,7 +51,7 @@ type, public :: client_type
 
   logical(kind=c_bool) :: cluster = .false.        !< True if a database cluster is being used
   type(c_ptr)          :: client_ptr = c_null_ptr !< Pointer to the initialized SmartRedisClient
-
+  logical              :: is_initialized = .false.    !< True if client is initialized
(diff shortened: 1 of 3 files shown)


ashao (approved):
This checks out on my machine and is a useful safety check to avoid instantiating multiple instances of the client unnecessarily. This is also needed to support the use of SmartRedis in both the NUOPC driver and standalone MOM6.

Could one of @al-rigazzi , @Spartee , @mellis13 check to ensure that it works on your internal machines as well?

mellis13:
@ashao I'll take a look and make sure this passes on our internal machines

billschereriii (changes requested):
Change looks reasonable and implementation is correct. However, should add a test case to attempt duplicate initialization and to validate that is_initialized is returning what we expect it to

jedwards4b:
@billschereriii thank you for the review, I have added a test as requested.

billschereriii:
Update looks good to me

billschereriii (approved):

billschereriii:
merged

PR #95, opened 2021-07-20
Fortran cleanup

billschereriii:
Assorted code cleanup

1. Add intents to all API functions
2. Unify string declaration and handling
 * Standardize declarations
 * Make sure Transfer and Trim are used appropriately and consistently
3. Rename "this" variable to something more intuitive for Fortran
4. Fix memory leaks when calling convert_char_array_to_c

### DIFF
--- a/src/fortran/client.F90
+++ b/src/fortran/client.F90
@@ -610,8 +610,10 @@ subroutine set_model( client, key, model, backend, device, batch_size, min_batch
 
   deallocate(c_inputs)
   deallocate(input_lengths)
+  !deallocate(ptrs_to_inputs)
   deallocate(c_outputs)
   deallocate(output_lengths)
+  !deallocate(ptrs_to_outputs)
 end subroutine set_model
 
 !> Execute a model
@@ -627,7 +629,7 @@ subroutine run_model(client, key, inputs, outputs)
 
   integer(c_size_t), dimension(:), allocatable, target :: input_lengths, output_lengths
   integer(kind=c_size_t) :: n_inputs, n_outputs, key_length
-  type(c_ptr)            :: inputs_ptr, input_lengths_ptr, outputs_ptr, output_lengths_ptr
+  type(c_ptr) :: inputs_ptr, input_lengths_ptr, outputs_ptr, output_lengths_ptr
   type(c_ptr), dimension(:), allocatable :: ptrs_to_inputs, ptrs_to_outputs
 
   integer :: i
@@ -646,16 +648,17 @@ subroutine run_model(client, key, inputs, outputs)
 
   deallocate(c_inputs)
   deallocate(input_lengths)
+  !deallocate(ptrs_to_inputs)
   deallocate(c_outputs)
   deallocate(output_lengths)
-
+  !deallocate(ptrs_to_outputs)
 end subroutine run_model
 
 !> Retrieve the script from the database
 subroutine get_script(client, key, script)
-  class(client_type),               intent(in   ) :: clien  !< An initialized SmartRedis client
-  character(len=*),                 intent(in   ) :: key    !< The key to use to place the script
-  character(len=*),                 intent(  out) :: script !< The script as a continuous buffer
+  class(client_type), intent(in   ) :: client !< An initialized SmartRedis client
+  character(len=*),   intent(in   ) :: key    !< The key to use to place the script
+  character(len=*),   intent(  out) :: script !< The script as a continuous buffer
 
   ! Local variables
   character(kind=c_char,len=len_trim(key)) :: c_key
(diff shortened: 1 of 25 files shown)


billschereriii:
All items addressed; all tests pass

ashao:
Nice catch on the deallocates and thanks for the general cleanup! I think the only point of discussion from my side is the refactor of the `this`. I know it's a pretty soft convention, but I partly prefer a more generic variable for the object to which the procedure is bound so that:

1. It can immediately be recognized as a type-bound procedure
2. Avoid false positives if searching through function/subroutine signatures to find non-type bound procedures.

The way that I think of it is the equivalent of `self` in Python. What do you think about `self` instead of `this` to make it more broadly readable?

ashao:
As a side note, I really wish the Fortran standards committee would actually codify something like `this` or `self` or some actual keyword when defining a type-bound procedure!

billschereriii:
Thanks for the review feedback Andrew.  I can live with `self` instead of `this`, so I'll go ahead and make that change in the next few days. Was there anything else you were concerned about?

ashao (changes requested):
I agree with the functional changes introduced by this PR. The explicit intents that were added (except where indicated in one location) are appropriate.

As was added to the discussion on the PR: it seems to be fairly common, but soft convention to use `this` to refer to the object on which a type-bound procedure is operating on. I personally dislike `this`, but do find some kind of generic name to be more useful to be able to distinguish a type-bound procedure from a more generic subroutine. As we had discussed in the PR thread, `self` would be fine with me and follows the Python convention.

After you make these changes, I'll test locally and confirm that it passes.

src/fortran/client.F90
```diff
@@ -135,74 +135,74 @@ module smartredis_client
...
 
+  if (present(cluster)) client%cluster = cluster
+  client%client_ptr = c_constructor(client%cluster)
 end subroutine initialize
 
 !> A destructor for the SmartRedis client
-subroutine destructor( this )
-  class(client_type) :: this
+subroutine destructor( client )
+  class(client_type), intent(in) :: client
```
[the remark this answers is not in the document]
billschereriii: Agree 100% with both suggestions. Done

billschereriii:
All requested updates complete. Passes Fortran tests on Horizon

src/fortran/client.F90
```diff
@@ -136,24 +136,24 @@ module smartredis_client
 
 !> Initializes a new instance of a SmartRedis client
 subroutine initialize( client, cluster )
```
ashao, line 138: This should also be `self`

src/fortran/dataset.F90
```diff
@@ -87,231 +87,261 @@ module smartredis_dataset
 
 contains
 
-subroutine initialize( this, name )
-  class(dataset_type) :: this
-  character(len=*)    :: name !< Name of the dataset
+subroutine initialize( dataset, name )
```
ashao, line 90: Also should be `self`

ashao (changes requested):
Looks good! All I can see is a couple places where it's still `client` or `dataset` instead of `self`

src/fortran/client.F90
billschereriii: This is the constructor, so I thought it merited special treatment. Do you disagree?

src/fortran/dataset.F90
billschereriii: This is the constructor, so I thought it merited special treatment. Do you disagree?

src/fortran/client.F90
ashao: Could you elaborate as to why you think it would warrant a deviation? The main one that I can think of is that it'd be easier to grep for the initialization routines for a given type. From my point of view is that these initialize routines are still a type-bound procedure and so it seems more consistent to me that all such procedures should use `self`. 

Maybe to satisfy both we could rename the subroutine name and signature to be
```
 subroutine initialize_client(self, cluster)
```

and then in the definition of the type:

```
type, public :: client_type

[...]

contains

procedure, public :: initialize => initialize_client

end type client_type
```

Thoughts?

billschereriii: Works for me!

billschereriii: Works for me. I'll do something similar for dataset

billschereriii: Changes complete, review requested

src/fortran/client.F90
```diff
@@ -63,7 +63,7 @@ module smartredis_client
                               unpack_tensor_float, unpack_tensor_double
 
   !> Initializes a new instance of the SmartRedis client
-  procedure :: initialize
+  procedure :: initialize_client
```
ashao, line 66: Sorry I think I was a little unclear. I meant that here we would map the procedure `initialize` to `initialize_client.` So that we can still do something like
```
call client_instance%initialize()
```
as opposed to
```
call client_instance%initialize_client()
```
which seems a bit verbose.

So change this line to:
```
  procedure :: initialize => initialize_client
```

src/fortran/dataset.F90
```diff
@@ -45,7 +45,7 @@ module smartredis_dataset
   contains
 
   !> Initialize a new dataset with a given name
-  procedure :: initialize
+  procedure :: initialize_dataset
```
ashao, line 48: Similar treatment to the initialize subroutine for `client_type`

ashao (changes requested):
Requesting a small refactor of the type bound initializers for `client` and `dataset`.

src/fortran/client.F90
billschereriii: Thanks for the suggestion. It's implemented and ready for review. I agree that it makes the code better to read

ashao: Great thanks @billschereriii!

ashao (approved):
Thanks for iterating with me on this @billschereriii. It looks like bringing in Jim Edwards' PR caused a conflict which looks like it should be pretty simple to resolve. Other than that, it's good to go.

By the way, I somehow borked my local SmartRedis testing environment, but considering that all the subsequent changes to your initial tests are cosmetic, I doubt anything will have broken. Might be good to do one last test on horizon depending on the paranoia levels.

billschereriii:
Thanks for the good feedback.  I resolved the merge conflict and will double-check that it still works before doing a final branch merge.

I've been testing each set of changes as I've gone, so I can assure you that this will pass the unit tests before I merge the branch.

billschereriii:
Final version approved by Andrew, with changes from isinitialized() addition merged in

Issue #92, opened 2021-07-19
Failed connection error not raised to Python


mellis13:
# Description
If the C++ client fails to connects to a database, the failed connection is not raised as a ``std::runtime_error`` to the Python client.  This should be changed as all other errors get raised as ``std::runtime_error`` to the Python client.

# How to reproduce

Attempt to connect the Python client to a nonexistent database.

# Expected behavior

A ``std::runtime_error`` should be raised.

# System
All

PR #97, opened 2021-07-22
Improve error handling for client initialization

mellis13:
This PR addresses #92 by raising all errors encountered during client creation and initialization as ``std::runtime_error``.  Additionally, a client connection is forced in the non-cluster client creation by running the Redis ``ping`` command so that a client connection is made.  This improves error tracking by having the source of a bad client connection traceable to ``_connect()``.

### DIFF
--- a/src/cpp/redis.cpp
+++ b/src/cpp/redis.cpp
@@ -61,6 +61,7 @@ CommandReply Redis::run(Command& cmd)
     while (n_trials > 0 && success) {
 
         try {
+            std::cout<<"Trying command execution "<<std::endl;
             reply = this->_redis->command(cmd_fields_start, cmd_fields_end);
 
             if(reply.has_error()==0)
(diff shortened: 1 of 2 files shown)


src/cpp/redis.cpp
```diff
@@ -329,29 +329,50 @@ CommandReply Redis::get_script(const std::string& key)
...
-                      << " more trials will be made.";
-          n_connection_trials--;
-          std::this_thread::sleep_for(std::chrono::seconds(2));
+        catch (sw::redis::Error& e) {
+            if(i == n_trials) {
+                delete this->_redis;
+                this->_redis = 0;
+                throw std::runtime_error(e.what());
+            }
+            std::this_thread::sleep_for(std::chrono::seconds(2));
```
billschereriii, line 369: Sleeping from inside a catch block is dangerous -- if the sleep call throws an exception, there is no way to catch it and the app shuts down immediately.

billschereriii: Just set a failed boolean and then do the entire if block outside the exception handler

src/cpp/rediscluster.cpp
```diff
@@ -467,28 +467,30 @@ CommandReply RedisCluster::get_script(const std::string& key)
...
+        catch (std::bad_alloc& e) {
+            throw std::runtime_error(e.what());
+        }
+        catch (sw::redis::Error& e) {
+            delete this->_redis_cluster;
+            this->_redis_cluster = 0;
+            if(i == n_trials) {
+                throw std::runtime_error(e.what());
+            }
+            std::this_thread::sleep_for(std::chrono::seconds(2));
```
billschereriii, line 486: Same as above -- do this block outside the exception handler

billschereriii (changes requested):
See comments attached to source files  (basically, never execute code that can throw exceptions from inside exception handlers)

Also, be sure to put a try/catch block around the call to wait_for()

src/cpp/redis.cpp
mellis13: Yeah, I can move it outside if we wanted to be extra safe, but ``sleep_for`` has a no-throw guarantee when using durations defined in std::chrono.

PR #98, opened 2021-07-26
Change error output to just emit the first field

mellis13:
A failed command execution will now emit only the first field in order to prevent large amounts of data being printed.


billschereriii (approved):
Looks reasonable, though at some point we may want to dump the lot for debug purposes. I'll eventually add a debug_verbosity_level setting to the code; we can add this then.

PR #99, opened 2021-07-27
C interface cleanup

billschereriii:
This pull addresses:
 * C cleanup (Poseidon ticket 61)
 * C stylistics (Poseidon ticket 59)
 * Const correctness and consistency for interface parameters (discovered in the process of this effort)
 * Inability to cleanly deallocate dataset objects
 * Compiler warnings from missing default cases in switch statements (triggered by adding `undefined` enum value to `TensorType`)


CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ option(BUILD_PYTHON "Build the python module" ON)
 
 set(CMAKE_BUILD_TYPE RELEASE)
 set(CMAKE_CXX_STANDARD 17)
+#SET(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -Wall -Werror")
```
mellis13, line 9: Was this meant to be left as a comment?  If so, do we still want to leave it in?

include/c_dataset.h
```diff
@@ -49,6 +50,13 @@ extern "C" {
 void* CDataSet(const char* name,
                const size_t name_length);
 
+/*!
+*   \brief C-DataSet destructor
+*   \param name The name of the dataset
+*   \param dataset A c-ptr to the dataset object
+*/
+void DeallocateeDataSet(void* dataset);
```
mellis13, line 58: In the  c client interface we use ``DeleteCClient``.  Should we call this ``DeleteDataSet`` to be consistent?

include/c_dataset.h
```diff
@@ -49,6 +50,13 @@ extern "C" {
 void* CDataSet(const char* name,
                const size_t name_length);
 
+/*!
+*   \brief C-DataSet destructor
+*   \param name The name of the dataset
```
mellis13, line 55: ``name `` isn't an input parameter for this function.

include/c_dataset.h
billschereriii: There is already a delete_dataset function that removes the dataset from Redis, so I wanted to try to avoid confusion

CMakeLists.txt
billschereriii: I meant to remove it, thanks

include/c_dataset.h
billschereriii: Good catch! I'll fix that

src/c/c_dataset.cpp
```diff
@@ -30,163 +30,196 @@
...
-  std::string name_str = std::string(name, name_length);
-  DataSet* dataset = new DataSet(name_str);
-  return (void*)dataset;
+  // Sanity check parameters
+  if (NULL == name)
+    return NULL;
+
+  std::string name_str(name, name_length);
+
+  DataSet *dataset = NULL;
```
mellis13, line 44: I know in pure c programs the convention is usually that the ``*`` is attached to the variable name, but the convention in SmartRedis has been to follow C++ norms (ref: https://www.stroustrup.com/bs_faq2.html#whitespace)  and to attach the ``*`` to the type.  I can see us potentially wanting to have the c-style function prototypes reflect c-standards, but I think internal code (especially code in ``.cpp`` files), should follow C++ conventions.

src/c/c_dataset.cpp
```diff
@@ -30,163 +30,196 @@
...
+  
+  return reinterpret_cast<void *>(dataset);
 }
 
+// Deallocate a DataSet
+extern "C"
+void DeallocateeDataSet(void *dataset)
+{
+  // Sanity check parameters
+  if (NULL == dataset)
```
mellis13, line 59: Maybe this can be condense to remove mid-function ``return`` statement with:

```
 if (NULL != dataset)
           delete reinterpret_cast<DataSet *>(dataset)
```

src/c/c_dataset.cpp
```diff
@@ -30,163 +30,196 @@
...
-  This function may allocate new memory for the tensor.
-  This memory will be deleted when the user deletes the
-  DataSet object.
-  */
-  DataSet* d = (DataSet*)dataset;
-  std::string name_str = std::string(name, name_length);
-
-  TensorType t_type;
+  // Sanity check parameters
+  if (NULL == dataset || NULL == name || NULL == data ||
```
mellis13, line 142: Can't ``data`` be NULL since we are allocating the memory for them?  It seems totally fine for user to pass in ``NULL``.

src/c/c_dataset.cpp
```diff
@@ -30,163 +30,196 @@
...
+// memory that is cleared when the user deletes the
+// DataSet object
 extern "C"
-void* get_meta_scalars(void* dataset,
-                      const char* name,
-                      const size_t name_length,
-                      size_t* length,
-                      CMetaDataType* type)
+void *get_meta_scalars(void *dataset,
+		       const char *name,
```
mellis13, line 186: Maybe fix alignment so that all types are in vertical alignment.

src/cpp/client.cpp
```diff
@@ -223,6 +223,8 @@ void Client::put_tensor(const std::string& key,
             tensor = new Tensor<uint8_t>(p_key, data, dims,
                                          type, mem_layout);
             break;
+        default :
+	    return; // FINDME: Need to handle this case better
```
mellis13, line 227: For these default cases in the C++ client, should we throw ``std::runtime_error`` to be consistent with all of the other invalid inputs or cases throughout the code?

src/fortran/client.F90
```diff
@@ -135,74 +135,74 @@ module smartredis_client
...
+  logical, optional,  intent(in   ) :: cluster !< If true, client uses a database cluster (Default: .false.)
 
+  if (present(cluster)) client%cluster = cluster
+  client%client_ptr = c_constructor(client%cluster)
 end subroutine initialize
 
 !> A destructor for the SmartRedis client
-subroutine destructor( this )
-  class(client_type) :: this
+subroutine destructor( client )
```
mellis13, line 147: I thought ``client.F90`` changes were in a separate PR?  Did something get messed up on merge?

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
   s->put_dataset(*d);
-  return;
 }
 
+// Return a pointer to a new dataset.  The user is
+// responsible for deleting the dataset via DeallocateeDataSet()
 extern "C"
-void* get_dataset(void* c_client, const char* name,
-                  const size_t name_length)
+void *get_dataset(void *c_client, const char *name, const size_t name_length)
```
mellis13, line 79: Split up long line

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
-  /* Return a pointer to a new dataset.  The user is
-  responsible for deleting the dataset.
-  */
-  Client* s = (Client *)c_client;
-  std::string dataset_name = std::string(name, name_length);
-  DataSet* dataset = new DataSet(s->get_dataset(dataset_name));
-  return (void*)dataset;
+  // Sanity check params
+  if (NULL == c_client || NULL == name)
+    return NULL; // Nothing we can do, so just bail. (error reporting to come later)
```
mellis13, line 83: Put this comment above to limit long lines.

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
-  /* Return a pointer to a new Client.
-  The user is responsible for deleting the client.
-  */
-  Client* s = new Client(cluster);
-  return (void*)s;
+  Client *s = NULL;
+
+  try {
+    s = new Client(cluster);
+  } catch (...) {
```
mellis13, line 43: If we are going to return ``NULL`` on all exceptions, we should probably split this up because you can have memory leak if ``new`` operator succeeds but some redis error is thrown.  I would split into a ``bad_alloc`` catch and then have a ``...`` catch which would have a delete operation to prevent memory leaks.

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
 extern "C"
-void DeleteCClient(void* c_client)
+void DeleteCClient(void *c_client)
 {
-  /* This function frees the memory associated
-  with the c client.
-  */
-  Client* s = (Client*)c_client;
+  // Sanity check params
+  if (NULL == c_client)
```
mellis13, line 55: Probably would get rid of unnecessary mid-function return with a ``if (NULL != c_client)``

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
+  // Sanity check params
+  if (NULL == c_client || NULL == name)
+    return NULL; // Nothing we can do, so just bail. (error reporting to come later)
+
+  Client *s = reinterpret_cast<Client *>(c_client);
+  std::string dataset_name(name, name_length);
+  DataSet *dataset = NULL;
+
+  try {
+    dataset = new DataSet(s->get_dataset(dataset_name));
```
mellis13, line 90: Right now the only exception ``new DataSet`` can throw is probably ``bad_alloc`` because we aren't doing anything except assigning string name.  However, to be safe and prevent memory leaks, maybe we split this one up with a ``bad_alloc`` catch and a ``...`` catch like the ``new client`` call?

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
                     const size_t new_name_length)
 {
-  /* Rename a dataset in the database.
-  */
-  Client* s = (Client *)c_client;
-  std::string name_str = std::string(name, name_length);
-  std::string new_name_str = std::string(new_name, new_name_length);
+  // Sanity check params
+  if (NULL == c_client || NULL == name || NULL == new_name)
+    return; // Nothing we can do, so just bail. (error reporting to come later)
```
mellis13, line 105: Put comment above to limit long line.

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
+                  const size_t dest_name_length)
 {
-  /* Copy a dataset from teh src_name to the dest_name
-  */
-  Client* s = (Client *)c_client;
-  std::string src_name_str = std::string(src_name, src_name_length);
-  std::string dest_name_str = std::string(dest_name, dest_name_length);
+  // Sanity check params
+  if (NULL == c_client || NULL == src_name || NULL == dest_name)
+    return; // Nothing we can do, so just bail. (error reporting to come later)
```
mellis13, line 122: Same as before.  Comment above to limit long lines

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
-void delete_dataset(void* c_client, const char* name,
-                    const size_t name_length)
+void delete_dataset(void *c_client, const char *name, const size_t name_length)
 {
-  /* Delete a dataset (all metadata and tensors) from the
-  database.
-  */
-  Client* s = (Client *)c_client;
-  std::string dataset_name = std::string(name, name_length);
+  // Sanity check params
```
mellis13, line 135: I wonder if this ``// Sanity check params`` comment is needed throughout the code or we just remove.  It seems intuitive what is being checked.

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
+                const CTensorType type,
+                const CMemoryLayout mem_layout)
 {
-  /* Put a tensor of a specified type into the database
-  */
-  Client* s = (Client *)c_client;
-  std::string key_str = std::string(key, key_length);
+  // Sanity check params
+  if (NULL == c_client || NULL == key || NULL == data || NULL == dims)
+    return; // Nothing we can do, so just bail. (error reporting to come later)
```
mellis13, line 157: Thinking more about it, I think it would be more helpful to have this kind of behavior documented ahead of the function (in the header file) instead of in the function.

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
   }
 
   std::vector<std::string> output_vec;
-  if (n_outputs == 1 && output_lengths[0] == 0) {
-    output_vec = std::vector<std::string>();
-  }
-  else {
-    for(size_t i=0; i<n_outputs; i++) {
-      output_vec.push_back(std::string(outputs[i], output_lengths[i]));
+  if (1 != n_outputs || 0 != output_lengths[0]) {
```
mellis13, line 312: I know the advantage of the rvalue on the left side of ``==`` comparison helps guard against accidental assignment, but the isocpp core guidelines seem to just use the ``n_outputs != 1`` for readability.  We should talk as a group about this so that we can be consistent.  At the very least, it seems hard to switch between the two coding styles.

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
+  std::string function_str(function, function_length);
 
   std::vector<std::string> input_vec;
-  for(size_t i=0; i<n_inputs; i++)
-    input_vec.push_back(std::string(inputs[i], input_lengths[i]));
+  if (1 != n_inputs || 0 != input_lengths[0]) {
+    for (size_t i=0; i<n_inputs; i++) {
+      input_vec.push_back((NULL != inputs[i] && 0 != input_lengths[i]) ?
+			  std::string(inputs[i], input_lengths[i]) :
+			  std::string(""));
```
mellis13, line 488: This seems to implicitly allow for input tensor names that are empty.  is that ok?  It seems like that's something we shouldn't allow.

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
+    }
+  }
 
   std::vector<std::string> output_vec;
-  for(size_t i=0; i<n_outputs; i++)
-    output_vec.push_back(std::string(outputs[i], output_lengths[i]));
+  if (1 != n_outputs || 0 != output_lengths[0]) {
+    for (size_t i=0; i<n_outputs; i++) {
+      output_vec.push_back((NULL != outputs[i] && 0 != output_lengths[i]) ?
+			   std::string(outputs[i], output_lengths[i]) :
```
mellis13, line 496: This seems to implicitly allow for output tensor names that are empty.  is that ok?  It seems like that's something we shouldn't allow.

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
+  std::string name_str(name, name_length);
+
   return s->model_exists(name_str);
 }
 
+// Check whether a tensor exists in the database
 extern "C"
-bool tensor_exists(void* c_client, const char* name,
-                   const size_t name_length)
+bool tensor_exists(void *c_client, const char *name, const size_t name_length)
```
mellis13, line 577: Can we split these long lines up?

src/c/c_client.cpp
```diff
@@ -30,517 +30,654 @@
...
 }
 
+// Delay until a key exists in the database
 extern "C"
-bool poll_key(void* c_client,
-              const char* key, const size_t key_length,
-              const int poll_frequency_ms, const int num_tries)
+bool poll_key(void *c_client,
+              const char *key,
+	      const size_t key_length,
```
mellis13, line 593: Check that types are aligned vertically when possible.

src/c/c_dataset.cpp
billschereriii: I was thinking that the C interface should follow C style and the C++ interface could follow the new C++ style, but I'm open to discussion.

src/c/c_dataset.cpp
billschereriii: When every routine starts with a sanity check of this form, it becomes a sort of automatic checklist item. Consistency across the codebase helps. I'd rather not change this, but we can discuss

src/c/c_dataset.cpp
billschereriii: definitely. Not sure what happened here, probably a tab/space issue

src/cpp/client.cpp
billschereriii: I can do that as a temporary measure, though we're going to have to switch away from exceptions as our error handling scheme longer term if we want to properly support C and Fortran clients.  We need to have a group discussion on this topic after the release.

src/fortran/client.F90
billschereriii: I'm not sure how all the Fortran stuff got into this pull -- that's not even the final state. I'll back out all of these changes.

src/c/c_client.cpp
billschereriii: I have always used 77 characters as the target max length for a line. Is there a project standard here?

src/c/c_client.cpp
billschereriii: It's a placeholder for proper error handling, so guaranteed to go away in the next release. That's why I was willing to let it be an overlong line short term

src/c/c_client.cpp
billschereriii: Good point! I will split this.

src/c/c_client.cpp
billschereriii: See previous comment -- I like to keep the sanity checks separate

src/c/c_client.cpp
billschereriii: Agreed -- never know what can happen in the future

src/c/c_client.cpp
billschereriii: Need to chat on this one, per my response to your comment on the previous instance

src/c/c_client.cpp
billschereriii: Whatever we agree to I'll do universally

src/c/c_client.cpp
billschereriii: Extra comments don't hurt. I'd rather see more comments in the code rather than less.

src/c/c_client.cpp
billschereriii: Not clear what you're asking for here

src/c/c_client.cpp
billschereriii: Agree we should have a group discussion on this

src/c/c_client.cpp
billschereriii: Better than passing a NULL and blowing up Redis. I welcome suggestions for better handling of this case

src/c/c_client.cpp
billschereriii: See previous

src/c/c_client.cpp
billschereriii: See previous comment on line lengths

src/c/c_dataset.cpp
mellis13: Yeah, I understand that reasoning.  I think given that this is C++ code though (uses standard containers and all) we should follow C++ standards.  Also, helps to have few conventions in the code.

src/c/c_client.cpp
billschereriii: probably a tab/space issue from my editor. Will fix.

src/c/c_dataset.cpp
billschereriii: I can see that. Consider it done.

src/c/c_client.cpp
mellis13: I didn't actually check character count (it just looked long).  We've been doing <79 characters, so if it is <79 ignore the comment.

src/c/c_client.cpp
mellis13: Sorry this wasn't clear.  Where appropriate I was thinking it would be helpful to have this temporary error handling documented (no action error handling or returning NULL pointer, etc) in the doc strings in the header files.  But since this is temporary, it's not a big deal at all.

src/c/c_client.cpp
mellis13: If we are going to do the check and pass in an empty string, maybe we should do the check separately and not execute the command in C client?  That would follow the pattern you've been using on no action for improper inputs.

src/c/c_client.cpp
billschereriii: TODO: Document NULL return in headers

src/c/c_client.cpp
billschereriii: Done

src/c/c_client.cpp
billschereriii: Per group discussion, tabling use of Yoda conditionals for now

billschereriii:
All feedback addressed and all tests pass. Ready for re-review

src/c/c_client.cpp
```diff
@@ -245,42 +289,59 @@ void set_model_from_file(void* c_client,
...
+  // that the array should be skipped) but if more than one entry is present, the
+  // strings must be nonzero length
+  if (n_inputs != 1 && input_lengths[0] != 0) {
+    for (size_t i = 0; i < n_inputs; i++){ 
+      if (inputs[i] == NULL || input_lengths[i] == 0) {
+        return; // Bad param = bail
+      }
+    }
+  }
+  if (n_outputs > 0 != 1 && output_lengths[0] != 0) {
```
mellis13, line 309: Should this just be ``n_outputs != 1``?

mellis13 (approved):
Looks good to merge.  Thanks for the improvements!

PR #100, opened 2021-07-27
Dataset exists

billschereriii:
Address Poseidon ticket 88.
Added dataset_exists command and unit testing for c, cpp, fortran, python.
Minor incompatibility with c cleanup pr -- will need to clean up afterwards

### DIFF
--- a/include/c_client.h
+++ b/include/c_client.h
@@ -586,8 +586,8 @@ bool model_exists(void* c_client,
 *   \returns Returns true if the key exists in the database
 */
 bool dataset_exists(void* c_client,
-                   const char* name,
-                   const size_t name_length);
+                    const char* name,
+                    const size_t name_length);
 
 /*!
 *   \brief Check if the key exists in the database at a
--- a/include/c_client.h
+++ b/include/c_client.h
@@ -542,7 +542,7 @@ bool key_exists(void* c_client,
                 const size_t key_length);
 
 /*!
-*   \brief Check if a tensor or dataset exists in the database
+*   \brief Check if a tensor exists in the database
 *   \param c_client A pointer to c client
 *                   to use for communication
 *   \param name The name of the entity that will be checked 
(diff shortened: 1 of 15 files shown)


include/c_client.h
```diff
@@ -573,6 +573,22 @@ bool model_exists(void* c_client,
                   const char* name,
                   const size_t name_length);
 
+  /*!
+*   \brief Check if a dataset exists in the database
+*   \param c_client A pointer to c client
+*                   to use for communication
+*   \param name The name of the entity that will be checked 
+*               in the database. The full key associated to
+*               \p name will formed according to the
```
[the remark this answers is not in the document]
billschereriii: Fixed 4 or 5 instances

include/c_client.h
```diff
@@ -573,6 +573,22 @@ bool model_exists(void* c_client,
...
+*               prefixing behavior
+*   \param name_length The length of the name c-string,
+*                      excluding null terminating character
+*   \returns Returns true if the key exists in the database
+*/
+bool dataset_exists(void* c_client,
+                   const char* name,
```
mellis13, line 589: Align ``const`` with ``void*``.  Same with line below.

include/client.h
```diff
@@ -441,15 +441,25 @@ class Client
...
         *               Depending on the current prefixing behavior,
         *               the name could be automatically prefixed
         *               to form the corresponding key.
-        *   \returns Returns true if the key exists in the database
+        *   \returns Returns true if the tensor key exists in the database
         */
         bool tensor_exists(const std::string& name);
 
+        /*!
+        *   \brief Check if the ataset exists in the database
```
mellis13, line 454: ``ataset`` -> ``dataset``

include/c_client.h
billschereriii: fixed

include/client.h
billschereriii: fixed

src/cpp/client.cpp
```diff
@@ -572,6 +572,13 @@ bool Client::tensor_exists(const std::string& name)
     return this->_redis_server->key_exists(g_key);
 }
 
+bool Client::dataset_exists(const std::string& name)
+{
+    // Same implementation as for tensors; the next line is NOT a type
+    std::string g_key = this->_build_tensor_key(name, true);
```
mellis13, line 578: We should probably use the ``_build_dataset_ack_key`` if we stick with the the current paradigm (see comment below for potential discussion).

mellis13 (approved):
All tests pass.  I'm inclined to create a separate issue for the ack key change and just change the one function call for now (_build_dataset_ack_key) and then merge.

src/cpp/client.cpp
```diff
@@ -572,6 +572,13 @@ bool Client::tensor_exists(const std::string& name)
...
+    return this->_redis_server->key_exists(g_key);
```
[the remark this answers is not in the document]
billschereriii: Update made as requested. _build_dataset_ack_key() currently just calls _build_tensor_key(), so there is no immediate functional change, but this will presumably change in the future.

PR #101, opened 2021-07-28
Upgrade RedisAI to 1.2.3.

mellis13:
Upgrade RedisAI to 1.2.3.

### DIFF
--- a/build-scripts/build-redisai-cpu.sh
+++ b/build-scripts/build-redisai-cpu.sh
@@ -1,11 +1,14 @@
 #!/bin/bash
 
+# get the number of processors
+NPROC=$(python -c "import multiprocessing as mp; print(mp.cpu_count())")
+
 #Install RedisAI
 if [[ -f ./RedisAI/install-cpu/redisai.so ]]; then
     echo "RedisAI CPU has already been downloaded and installed"
 else
     if [[ ! -d "./RedisAI" ]]; then
-        git clone --recursive https://github.com/RedisAI/RedisAI.git --branch v1.2.3 --depth=1 RedisAI
+        GIT_LFS_SKIP_SMUDGE=1 git clone --recursive https://github.com/RedisAI/RedisAI.git --branch v1.2.3 --depth=1 RedisAI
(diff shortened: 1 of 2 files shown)


Spartee (changes requested):
A couple of things.

- lets add ``GIT_LFS_SKIP_SMUDGE=1`` in front of the clone line.
- add ``WITH_TFLITE=0`` to the get_deps line and the make line 
- add ``WITH_UNIT_TESTS=0`` and ``-j $(nproc)`` to the make line.

mellis13:
Good ideas.  Just to make sure, is the current ``WITH_TFLITE=0`` and ``WITH_UNIT_TESTS=0`` incorrectly done?

Spartee:
My stale branch didn't have those, apologies they are right. But yes the other two as well.

build-scripts/build-redisai-cpu.sh
```diff
@@ -15,7 +18,7 @@ else
     echo "Downloading RedisAI CPU dependencies"
     CC=gcc CXX=g++ WITH_PT=1 WITH_TF=1 WITH_TFLITE=0 WITH_ORT=0 bash get_deps.sh cpu
     echo "Building RedisAI"
-    CC=gcc CXX=g++ GPU=0 WITH_PT=1 WITH_TF=1 WITH_TFLITE=0 WITH_ORT=0 WITH_UNIT_TESTS=0 make -C opt clean build
+    CC=gcc CXX=g++ GPU=0 WITH_PT=1 WITH_TF=1 WITH_TFLITE=0 WITH_ORT=0 WITH_UNIT_TESTS=0 make -j $NPROC -C opt clean build
```
Spartee, line 21: I believe we can just use ``$(nproc)`` here as some SmartRedis users might not be using Python. Right?

Speaking of this.. are we requiring Python in order to build because of the python based cmake? Can a user build without Python? Should they?

Spartee (approved):
1 comment, you can decide if that needs to be addressed. Otherwise this looks good to merge.

PR #105, opened 2021-07-29
Add method to retrieve cluster node information and update exception handling

mellis13:
This PR adds a method ``get_db_cluster_info`` to retrieve the reply for a ``CLUSTER INFO`` command that is directed at a specific host and port.  Also, exception handling for command execution and client connection are updated to throw a std::runtime_error for any exception, not just sw::redis exceptions.

### DIFF
--- a/src/cpp/redis.cpp
+++ b/src/cpp/redis.cpp
@@ -56,9 +56,8 @@ CommandReply Redis::run(Command& cmd)
     CommandReply reply;
 
     int n_trials = 100;
-    bool success = true;
 
-    while (n_trials > 0 && success) {
+    while (n_trials > 0) {
 
         try {
             reply = this->_redis->command(cmd_fields_start, cmd_fields_end);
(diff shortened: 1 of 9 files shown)


src/cpp/redis.cpp
```diff
@@ -68,24 +67,21 @@ CommandReply Redis::run(Command& cmd)
...
-        catch (sw::redis::IoError &e) {
+        catch (sw::redis::ClosedError &e) {
             n_trials--;
             std::this_thread::sleep_for(std::chrono::seconds(2));
         }
-        catch (...) {
+        catch (std::exception& e) {
             n_trials--;
-            throw;
+            throw std::runtime_error(e.what());
```
al-rigazzi, line 80: Does it make sense to decrement `n_trials` if we're going to `throw`? I guess it will do no harm, but I wanted to understand if there was some mechanism I don't understand

al-rigazzi (approved):
Looks good to merge, I only have a very minor comment.

src/cpp/redis.cpp
mellis13: That's a good catch, Al.  I will delete it.

PR #106, opened 2021-07-30
Enhance error handling for Python client

mellis13:
This PR adds additional error handling to support Python client connection.  Additionally, it fixes an error that was introduced in the previous error handling enhance PR that repeatedly forms a connection for redis cluster.

### DIFF
--- a/src/cpp/redis.cpp
+++ b/src/cpp/redis.cpp
@@ -63,9 +63,8 @@ CommandReply Redis::run(Command& cmd)
             reply = this->_redis->command(cmd_fields_start, cmd_fields_end);
 
             if(reply.has_error()==0)
-                n_trials = -1;
-            else
-                n_trials = 0;
+                return reply;
+            n_trials = 0;
         }
         catch (sw::redis::IoError &e) {
             n_trials--;
@@ -340,12 +339,7 @@ inline void Redis::_connect(std::string address_port)
     try {
         this->_redis = new sw::redis::Redis(address_port);
     }
-    catch (std::bad_alloc& e) {
-        throw std::runtime_error(e.what());
-    }
     catch (std::exception& e) {
-        delete this->_redis;
-        this->_redis = 0;
         throw std::runtime_error("Failed to create Redis "\
                                  "object with error: " +
                                  std::string(e.what()));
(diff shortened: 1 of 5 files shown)


src/cpp/redis.cpp
```diff
@@ -356,15 +361,24 @@ inline void Redis::_connect(std::string address_port)
...
                 break;
             }
+            else {
+                run_sleep = true;
+            }
         }
         catch (std::exception& e) {
             if(i == n_trials) {
-                delete this->_redis;
-                this->_redis = 0;
```
billschereriii, line 363: Leave in the assignment to NULL (and change from 0 to NULL)

src/cpp/redis.cpp
```diff
@@ -356,15 +361,24 @@ inline void Redis::_connect(std::string address_port)
...
             }
             run_sleep = true;
         }
+        catch (...) {
+            if(i == n_trials) {
+                throw std::runtime_error("A non-standard exception "\
+                                         "encountered during client "\
+                                         "connection.");
+            }
+            run_sleep = true;
```
billschereriii, line 380: Fail back to caller if we hit a weird exception?

src/cpp/rediscluster.cpp
```diff
@@ -470,18 +475,22 @@ inline void RedisCluster::_connect(std::string address_port)
...
             this->_redis_cluster =
                 new sw::redis::RedisCluster(address_port);
-        }
-        catch (std::bad_alloc& e) {
-            throw std::runtime_error(e.what());
+            return;
         }
         catch (std::exception& e) {
-            delete this->_redis_cluster;
-            this->_redis_cluster = 0;
```
billschereriii, line 479: Leave in assignment to NULL (and switch 0 => NULL)

billschereriii:
PR looks good. A few minor comments above

src/cpp/redis.cpp
mellis13: I see what you are saying, but I think we should put the assignment to NULL up in teh construction portion, not the PONG portion

src/cpp/rediscluster.cpp
mellis13: Agreed.

src/cpp/redis.cpp
billschereriii: Agree, keep assignment with the deletion

billschereriii (approved):

PR #109, opened 2021-08-03
Update docs for 0.2.0 release.

mellis13:
This PR updates the version number in documentation and config files for 0.2.0 release.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 0.2.0
 -----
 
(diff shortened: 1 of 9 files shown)


al-rigazzi (approved):
One minor note: I see that in `changelog.rst` the release date is July 29. Is that still OK? Otherwise LGTM

EricGustin:
There are no longer any warnings when docs are built on MacOS, so looks good to me.

PR #110, opened 2021-08-04
Change unit test to use setenv instead of putenv.

mellis13:
This PR updates the unit test for SSDB methods to use setenv instead of putenv for improved behavior on linux systems.


EricGustin (approved):
Looks good and is working on my end.

PR #111, opened 2021-08-04
Update the slurm utility script to bind database nodes

mellis13:
This PR updates the slurm utility script to bind database 
nodes to a specific address.  This prevents inconsistencies in
addresses retrieved with CLUSTER SLOTS when running the test suite.
Updated documentation was also added to better inform users
why inconsistent addresses can throw an error when trying to run
get_db_node_info() command.

### DIFF
--- a/utils/create_cluster/slurm_cluster.py
+++ b/utils/create_cluster/slurm_cluster.py
@@ -5,15 +5,10 @@ import os
 def get_ip_from_host(host):
     ping_out = ping_host(host)
     found = False
-    print(host)
-    print(ping_out)
-    print('here')
     # break loop when the hostname is found in
     # the ping output
     for item in ping_out.split():
-        print(item)
         if found:
-            print('Found!')
             return item.split("(")[1].split(")")[0]
         if item == host:
             found = True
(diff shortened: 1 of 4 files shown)


utils/create_cluster/slurm_cluster.py
```diff
@@ -2,6 +2,22 @@
 from os import environ
 import os
 
+def get_ip_from_host(host):
+    ping_out = ping_host(host)
+    found = False
+    print(host)
+    print(ping_out)
+    print('here')
```
billschereriii, line 10: This looks like temporary debugging code that should be removed

utils/create_cluster/slurm_cluster.py
```diff
@@ -2,6 +2,22 @@
...
+    # break loop when the hostname is found in
+    # the ping output
+    for item in ping_out.split():
+        print(item)
+        if found:
+            print('Found!')
```
billschereriii, line 16: This looks like temporary debugging code that should be removed

Spartee (approved):
This looks good to me, it does rely on a single formatted ``etc/hosts`` and certain node naming schemes but this is developer facing only and for now, this is more than fine.

PR #113, opened 2021-08-06
Refactor CPP client command structure

ctandon11:
Refactor the CPP client command structure to account for multiple command types.
Follow the refactor design outlined [here](https://hpe-my.sharepoint.com/:p:/p/celia_tandon/ES9AeBqRIzdMhCYScG7HbmkBgS1z0cJpWZUh-QAeELahuQ?e=pas55l). 

1. Make `Command` base class of inheritance 
  structure with `SingleKeyCommand`, `MultiKeyCommand`, `CompoundCommand`, 
  `AddressAtCommand`, and `AddressAnyCommand` subclasses. 
2. Refactor `RedisServer::run()`into a method with 5 polymorphic variants for 
 the 5 Command types. 
3. Refactor `run(CommandList)`to account for lists of different Command types. 
4. Remove `CommandReplyParser` by putting appropriate functions inside of Command objects.


billschereriii:
First thoughts:

1. Several files in the PR don't seem like they should be there:
* `utils/create_cluster/6379.conf`
* `utils/create_cluster/6379.log`
* `utils/create_cluster/6380.conf`
* `utils/create_cluster/6380.log`
* `utils/create_cluster/6381.conf`
* `utils/create_cluster/6381.log`

2. Plus, I'm not sure about:
 * `.vscode/settings.json`

3. The changelog `doc/changelog.rst` is showing merge conflict headers (e.g. `<<<<<<< HEAD`). These should be resolved.

4. The sub-classes of keyedcommand and nonkeyedcommand are in the same headers and source files as they are. They should be in separate headers and source files -- try to keep to a `one class: one file` rule.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
         Redis* _redis;
 
         /*!
-        *   \brief Execute a Command
-        *   \param cmd The Command to execute
+        *   \brief Execute a non-keyed Command that addresses
+        *          the given db node(s)
```
mellis13, line 593: Small change, but should this be "Execute a non-keyed Command that address a given node" since the NonKeyedCommand class only holds one address and port?  This assumes we are going to accept multiple address and ports at the client level and not the Command level.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \param cmd The non-keyed Command to execute
```
mellis13, line 594: Maybe make the comment "The AddressAtCommand to execute" so that the type shows up correctly.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \returns The CommandReply after execution
+        */
+        inline CommandReply _run(AddressAtCommand& cmd);
+
+        /*!
+        *   \brief Execute a non-keyed Command that addresses
+        *          any db node(s)
+        *   \param cmd The non-keyed Command to execute
```
mellis13, line 602: Maybe make the comment "The AddressAnyCommand to execute" so that the type shows up correctly.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \brief Execute a non-keyed Command that addresses
+        *          the given db node(s)
+        *   \param cmd The non-keyed Command to execute
+        *   \returns The CommandReply after execution
+        */
+        inline CommandReply _run(AddressAtCommand& cmd);
+
+        /*!
+        *   \brief Execute a non-keyed Command that addresses
+        *          any db node(s)
```
mellis13, line 601: Small change, but should this be "Execute a non-keyed Command that addresses any node" since the NonKeyedCommand class only holds one address and port?

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \param cmd The non-keyed Command to execute
+        *   \returns The CommandReply after execution
+        */
+        inline CommandReply _run(AddressAnyCommand& cmd);
+
+        /*!
+        *   \brief Execute a single-keyed command Command
```
mellis13, line 608: Maybe change to "Execute a SingleKeyCommand" so the type shows up correctly.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \param cmd The single-keyed Command to execute
```
mellis13, line 609: Change single-keyed Command to SingleKeyCommand so type gets linked in docs.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \returns The CommandReply after execution
+        */
+        inline CommandReply _run(SingleKeyCommand& cmd);
+
+        /*!
+        *   \brief Execute a multi-keyed command Command
```
mellis13, line 615: Change to "Execute a MultiKeyCommand" so that the type gets linked.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \param cmd The mulit-keyed Command to execute
```
mellis13, line 616: Change to MultiKeyCommand so the type gets linked in docs.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \returns The CommandReply after execution
+        */
+        inline CommandReply _run(MultiKeyCommand& cmd);
+
+        /*!
+        *   \brief Execute a keyed multi-command Command
```
mellis13, line 622: Update comment to CompoundCommand

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        *   \param cmd The keyed multi-command Command to execute
```
mellis13, line 623: Change multi-command Command to CompoundCommand so the type is linked in the docs.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
         *   \returns The CommandReply after execution
```
mellis13, line 624: General comment: we should discuss what a ``CompoundCommand`` should return.  Maybe a single CommandReply makes sense, but given in the general case that multiple ``Command`` are executed, we might need something more general.

include/client.h
```diff
@@ -591,19 +589,49 @@ class Client
...
+        inline CommandReply _run(CompoundCommand& cmd);
 
         /*!
         *   \brief Execute a list of commands
         *   \param cmds The CommandList to execute
         *   \returns The CommandReply from the last
         *            command execution
         */
-        CommandReply _run(CommandList& cmd_list);
+        inline CommandReply _run(CommandList& cmd_list);
```
mellis13, line 634: I wonder if we should change the return type for ``_run(CommandList& cmd_list)`` so that the caller has access to all of the responses.

include/command.h
```diff
@@ -88,6 +92,12 @@ class Command
         */
         ~Command();
 
+        /*!
+        *   \brief Run any type of Command on the RedisServer.
```
mellis13, line 96: Maybe change to "Run this Command on the RedisServer"

include/command.h
```diff
@@ -88,6 +92,12 @@ class Command
...
+        *   \param r A pointer to the RedisServer
+        */
+        virtual CommandReply runme(RedisServer * r);
```
mellis13, line 99: We usually do ``_`` for method names.  Should this be ``run_me``?  Also, we try to associated the ``*`` symbol with the type, so it should be ``RedisServer*``

include/commandlist.h
```diff
@@ -82,11 +82,10 @@ class CommandList
...
 
         /*!
-        *   \brief Dynamically allocate a new Command
-        *          and return a pointer to the new Command
-        *   \returns Pointer to a new Command
+        *   \brief Adds a new Command to the Command List
+        *   \param new_cmd Pointer to the new Command
         */
-        Command* add_command();
+        void add_command(Command* new_cmd);
```
mellis13, line 88: As we discussed, I think we should change this to a templated function so that the ``CommandList`` owns the memory associated with the ``Command``.  e.g.:

``` c++
template <class T>
T* add_command();
```

include/keyedcommand.h
```diff
@@ -0,0 +1,135 @@
...
+{
+    public:
+
+        /*!
+        *   \brief Return a copy of all Command keys
+        *   \returns std::vector of Command keys
+        */
+        std::vector<std::string> get_keys();
+
+        virtual CommandReply runme(RedisServer * r);
```
mellis13, line 58: The class comments seem to indicate this is an intermediary class -- should we make this pure virtual like ``Command`` class?

include/keyedcommand.h
```diff
@@ -0,0 +1,135 @@
...
+{
+    public:
+        virtual CommandReply runme(RedisServer *r);
+
+        /*!
+        *   \brief This function will fill a vector with the dimensions of
+        *          the tensor.
+        *   \param reply CommandReply from running "AI.TENSORGET"
+        *   \return std::vecotr<size_t> with dimensions of the tensor
+        *   We assume right now that the META reply is always
```
mellis13, line 106: This additional comment should go int a ``\details`` section if it is still relevant and needs to be included.

include/keyedcommand.h
```diff
@@ -0,0 +1,135 @@
...
+        *   in the same order and we can index base reply elements array;
+        */
+        std::vector<size_t> get_tensor_dims(CommandReply& reply);
+
+        /*!
+        *   \brief This function returns a string view of the data
+        *          tensor blob value
+        *   \param reply CommandReply from running "AI.TENSORGET"
+        *   \return string view of the data tensor blob value
+        *   We are going to assume right now that the meta data reply
```
mellis13, line 116: This additional comment should go int a ``\details`` section under ``\brief`` if it is still relevant and needs to be included.

include/keyedcommand.h
```diff
@@ -0,0 +1,135 @@
...
+        *   is always in the same order and we are going to just
+        *   index into the base reply.
+        */
+        std::string_view get_tensor_data_blob(CommandReply& reply);
+
+        /*!
+        *   \brief This function returns a string of the tensor type
+        *   \param reply CommandReply from running "AI.TENSORGET"
+        *   \return string of the tensor type
+        *   We are going to assume right now that the meta data reply
```
mellis13, line 126: This additional comment should go int a ``\details`` section under ``\brief`` if it is still relevant and needs to be included.

include/nonkeyedcommand.h
```diff
@@ -0,0 +1,144 @@
...
+/*!
+*   \brief The NonKeyedCommand intermediary class constructs Client
+*          commands without keys. These commands use db node addresses.
+*   \details The KeyedCommand class has multiple methods for dealing
+*            with non-keyed commands.
+*/
+class NonKeyedCommand : public Command
+{
+    public:
+        /*!
```
mellis13, line 52: Same comment as the ``KeyedCommand`` classes.  We should explicitly declare the constructor, destructor, move assignment, copy assignment, move constructor, and copy constructor for each class.

include/nonkeyedcommand.h
```diff
@@ -0,0 +1,144 @@
...
+
+        /*!
+        *   \brief Get port that command will be
+        *          to be executed on
+        *   \return uint16_t of port
+        *           if port hasn't been set, returns 0
+        */
+        uint16_t get_port();
+
+        virtual CommandReply runme(RedisServer * r) = 0;
```
mellis13, line 78: Add doc strings to this function.

include/nonkeyedcommand.h
```diff
@@ -0,0 +1,144 @@
...
+    public:
+        virtual CommandReply runme(RedisServer *r);
+
+        /*!
+        *   \brief Parse database node information from get_db_node_info()
+        *          into a nested unordered_map
+        *   \param info containing the database node information
+        *   \return parsed_reply_map containing the database node cluster information
+        */
+        parsed_reply_map parse_db_cluster_info(std::string info);
```
mellis13, line 139: I wonder if we should make these static member functions so we can call parsing routines without need to create the object, i.e. ``ClusterInfoCommand::parse_db_cluster_info()``

include/redis.h
```diff
@@ -88,15 +88,49 @@ class Redis : public RedisServer
...
+        *   \returns The CommandReply from the
+        *            command execution
+        */
+        virtual CommandReply run(CompoundCommand& cmd);
+
+        /*!
+        *   \brief Run a non-keyed Command that addresses
+        *          the given db node(s) on the server
+        *   \param cmd The non-keyed Command that
+        *              addresses the given db node(s)
```
mellis13, line 119: Maybe just make the doc strings singular ``db node`` since the object only has one address and port?

include/redis.h
```diff
@@ -88,15 +88,49 @@ class Redis : public RedisServer
...
+        *   \returns The CommandReply from the
+        *            command execution
+        */
+        virtual CommandReply run(AddressAtCommand& cmd);
+
+        /*!
+        *   \brief Run a non-keyed Command that addresses
+        *          any db node(s) on the server
```
mellis13, line 127: Same here.  ``db node`` should probably be singular.

include/redisserver.h
```diff
@@ -119,10 +154,11 @@ class RedisServer {
         /*!
         *   \brief Get a Tensor from the server
         *   \param key The name of the tensor to retrieve
+        *   \param cmd The command to run on the server
         *   \returns The CommandReply from the get tensor server
         *            command execution
         */
-        virtual CommandReply get_tensor(const std::string& key) = 0;
+        virtual CommandReply get_tensor(const std::string& key, GetTensorCommand& cmd) = 0;
```
mellis13, line 161: I'm not sure that we need to pass the ``GetTensorCommand`` type down to ``RedisServer`` from ``Client``.  I think we can probably just construct the ``GetTensorCommand`` inside of ``RedisServer::get_tensor()`` if we make the parsing function a static member variable (i.e. we won't need the object only the class at the ``Client`` level to parse the output).

src/cpp/client.cpp
```diff
@@ -89,15 +89,15 @@ DataSet Client::get_dataset(const std::string& name)
...
+        GetTensorCommand cmd;
         cmd.add_field("AI.TENSORGET");
         cmd.add_field(tensor_key, true);
         cmd.add_field("META");
         cmd.add_field("BLOB");
         reply = this->_run(cmd);
-        reply_dims = CommandReplyParser::get_tensor_dims(reply);
-        blob = CommandReplyParser::get_tensor_data_blob(reply);
-        type = CommandReplyParser::get_tensor_data_type(reply);
+        reply_dims = cmd.get_tensor_dims(reply);
```
mellis13, line 98: If we change to static member function, we won't need to create this ``GetTensorCommand``

src/cpp/client.cpp
```diff
@@ -797,7 +818,7 @@ inline void Client::_append_with_put_prefix(
 
 inline CommandReply Client::_get_dataset_metadata(const std::string& name)
 {
-    Command cmd;
+    CompoundCommand cmd;
```
mellis13, line 821: In general putting and getting a ``DataSet`` is a compound Command, but the ``Client`` current breaks it up into pieces, and this ``HGETALL`` is actually a ``SingleKeyCommand``

src/cpp/client.cpp
```diff
@@ -866,8 +887,8 @@ void Client::_append_dataset_metadata_commands(CommandList& cmd_list,
                                  "a DataSet into the database that "\
                                  "does not contain any fields or "\
                                  "tensors.");
-
-    Command* cmd = cmd_list.add_command();
+    CompoundCommand* cmd = new CompoundCommand();
```
mellis13, line 890: This ``HMSET`` command is better executed as a ``SingleKeyCommand``.

src/cpp/client.cpp
```diff
@@ -885,14 +906,14 @@ void Client::_append_dataset_tensor_commands(CommandList& cmd_list,
...
 
-    Command* cmd;
     while(it != it_end) {
         tensor = *it;
         tensor_key =
             this->_build_dataset_tensor_key(dataset.name,
                                             tensor->name(),
                                             false);
-        cmd = cmd_list.add_command();
+        CompoundCommand* cmd = new CompoundCommand();
```
mellis13, line 915: I think this should be a ``SingleKeyCommand``

src/cpp/client.cpp
```diff
@@ -910,7 +931,8 @@ void Client::_append_dataset_ack_command(CommandList& cmd_list,
     std::string dataset_ack_key =
         this->_build_dataset_ack_key(dataset.name, false);
 
-    Command* cmd = cmd_list.add_command();
+    CompoundCommand* cmd = new CompoundCommand();
```
mellis13, line 934: Also a ``SingleKeyCommand``?

src/cpp/client.cpp
```diff
@@ -939,17 +961,15 @@ void Client::_unpack_dataset_metadata(DataSet& dataset,
 TensorBase* Client::_get_tensorbase_obj(const std::string& name)
 {
     std::string g_key = this->_build_tensor_key(name, true);
-
-    CommandReply reply = this->_redis_server->get_tensor(g_key);
+    GetTensorCommand cmd;
```
mellis13, line 964: I think if we put the parser functions as static member functions, we won't need to create this command.

src/cpp/command.cpp
```diff
@@ -80,6 +80,12 @@ Command::~Command()
     make_empty();
 }
 
+CommandReply Command::runme(RedisServer * r)
```
mellis13, line 83: If we make this a pure virtual function, we can remove this.

src/cpp/keyedcommand.cpp
```diff
@@ -0,0 +1,93 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "keyedcommand.h"
+#include "redisserver.h"
+
+using namespace SmartRedis;
+
+CommandReply KeyedCommand::runme(RedisServer * r)
```
mellis13, line 34: Do we need a ``runme`` in ``KeyedCommand``?  Should this be a pure virtual function?

src/cpp/keyedcommand.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+    CommandReply reply;
+    return reply;
+}
+
+CommandReply MultiKeyCommand::runme(RedisServer * r)
```
mellis13, line 40: Associate the ``*`` with the ``RedisServer`` type for all function prototypes in keyedcommand.cpp

src/cpp/nonkeyedcommand.cpp
```diff
@@ -26,81 +26,46 @@
...
-        throw std::runtime_error("The message does not have the correct "\
-                                "number of fields");
+CommandReply AddressAnyCommand::runme(RedisServer * r)
+{
+    return r->run(*this);
+}
 
-    return TENSOR_TYPE_MAP.at(std::string(reply[1].str(),
-                                            reply[1].str_len()));
+CommandReply DBInfoCommand::runme(RedisServer *r)
```
mellis13, line 62: For ``DBInfoCommand`` and ``ClusterInfoCommand`` do we need to implement ``runme``?  It seems like the execution would be the same as the parent class, so we can just rely on the ``virtual`` parent class definition?

src/cpp/redis.cpp
```diff
@@ -49,50 +49,29 @@ Redis::~Redis()
...
-    Command::iterator cmd_fields_start = cmd.begin();
-    Command::iterator cmd_fields_end = cmd.end();
-    CommandReply reply;
-
-    int n_trials = 100;
-
-    while (n_trials > 0) {
+CommandReply Redis::run(SingleKeyCommand& cmd){
+    Command* bcmd = &cmd;
+    return _run(*bcmd);
```
mellis13, line 54: ``_run`` should be ``this->_run``.  Also, do we need to create the pointer and then dereference the pointer?  Is there an easier to read and elegant way to call a common ``_run`` function for all of these ``run`` fuctions?

src/cpp/redis.cpp
```diff
@@ -101,7 +80,7 @@ CommandReply Redis::run(CommandList& cmds)
     CommandList::iterator cmd_end = cmds.end();
     CommandReply reply;
     while(cmd != cmd_end) {
-        reply = this->run(**cmd);
+        reply = ((Command *)(*cmd))->runme(this);
```
mellis13, line 83: We've decided to move away from c-style casts to c++ casts, so we should change ``(Command *)`` to the appropriate cast.  I also wonder if a private ``Redis`` function like ``Redis::_run_base_command(Command& cmd)`` would make this easier so that all of the methods could just call that without having to do the pointer creation and dereferencing seen in functions above.

src/cpp/redis.cpp
```diff
@@ -142,9 +120,8 @@ CommandReply Redis::put_tensor(TensorBase& tensor)
     return this->run(cmd);
 }
 
-CommandReply Redis::get_tensor(const std::string& key)
+CommandReply Redis::get_tensor(const std::string& key, GetTensorCommand& cmd)
```
mellis13, line 123: Same as before, we probably don't need to pass through the ``GetTensorCommand``.

src/cpp/rediscluster.cpp
```diff
@@ -51,64 +53,67 @@ RedisCluster::~RedisCluster()
...
-        db_prefix = this->_address_node_map.at(cmd.get_address() + ":"
-                    + std::to_string(cmd.get_port()))->prefix;
-    else if (cmd.has_keys())
+    if (cmd.has_keys())
         db_prefix = this->_get_db_node_prefix(cmd);
     else
         throw std::runtime_error("Redis has failed to find database");
 
-    std::string_view sv_prefix(db_prefix.data(), db_prefix.size());
+    Command *bcmd = &cmd;
```
mellis13, line 64: Same comment as ``Redis`` for this pointer creation and dereferencing.

src/cpp/rediscluster.cpp
```diff
@@ -179,7 +183,7 @@ CommandReply RedisCluster::rename_tensor(const std::string& key,
     uint16_t new_key_hash_slot = this->_get_hash_slot(new_key);
 
     if(key_hash_slot == new_key_hash_slot) {
-        Command cmd;
+        MultiKeyCommand cmd;
```
mellis13, line 186: ``rename_tensor`` is actually a ``CompoundCommand`` composed of a get and set for rediscluster.

doc/clients/fortran.rst
```diff
@@ -134,7 +134,12 @@ duplicated code and organize the variety of methods included within each type. T
 general functionality and suffixed by the type of code contained within.
 
 * ``<functionality>_interfaces.inc``: Define the C-bound interfaces to the SmartRedis-C methods
+<<<<<<< HEAD
+* ``<functionality>_methods_common.inc``: Represents the source code that is exactly the same for all methods which
+share the same functionality, but differ only by the type of variable
+=======
 * ``<functionality>_methods_common.inc``: Represents the source code that is exactly the same for all methods which share the same functionality, but differ only by the type of variable
+>>>>>>> 8efe7293bc6eb1722ca047a75ed5ef99d21ef789
```
billschereriii, line 142: Need to finish this merge

.vscode/settings.json
```diff
@@ -0,0 +1,97 @@
...
+        "*.tcc": "cpp",
+        "memory_resource": "cpp",
+        "netfwd": "cpp",
+        "stop_token": "cpp",
+        "cinttypes": "cpp",
+        "any": "cpp",
+        "cfenv": "cpp",
+        "regex": "cpp"
+    }
+}
```
billschereriii, line 97: This file shouldn't be checked in

include/clusterinfocommand.h
```diff
@@ -0,0 +1,57 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SMARTREDIS_CLUSTERINFOCOMMAND_H
+#define SMARTREDIS_CLUSTERINFOCOMMAND_H
+
+#include "addressatcommand.h"
+
+using parsed_reply_map = std::unordered_map<std::string, std::string>;
+using parsed_reply_nested_map = std::unordered_map<std::string, parsed_reply_map>;
```
billschereriii, line 35: If we don't use this type in this file it shouldn't be defined here.

include/commandlist.h
```diff
@@ -82,11 +82,11 @@ class CommandList
...
 
         /*!
-        *   \brief Dynamically allocate a new Command
-        *          and return a pointer to the new Command
-        *   \returns Pointer to a new Command
+        *   \brief Add new Commmand of type C to the CommandList
+        *   \tparam C Any type of Command
         */
-        Command* add_command();
+        template <class C>
```
billschereriii, line 88: Use T, not C, for template class to follow existing conventions

include/commandlist.tcc
```diff
@@ -0,0 +1,40 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SMARTREDIS_COMMANDLIST_TCC
+#define SMARTREDIS_COMMANDLIST_TCC
+
+template <class C>
+C* CommandList::add_command()
```
billschereriii, line 33: Use T, not C, for template class

include/dbinfocommand.h
```diff
@@ -0,0 +1,57 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SMARTREDIS_DBINFOCOMMAND_H
+#define SMARTREDIS_DBINFOCOMMAND_H
+
+#include "addressatcommand.h"
+
+using parsed_reply_map = std::unordered_map<std::string, std::string>;
```
billschereriii, line 34: If we aren't using this type in this file, don't define it here

include/keyedcommand.h
```diff
@@ -0,0 +1,108 @@
...
+        *            are copied using the copy assignment
+        *            operator for the derived type. This is meant
+        *            to provide functionality to deep
+        *            copy a Command.
+        *   \returns A pointer to dynamically allocated
+        *            derived type cast to parent Command
+        *            type.
+        */
+        virtual Command* clone() = 0;
+
```
billschereriii, line 98: It's a little odd to have a pure virtual function that overloads a defined function.
If command.cpp implements clone(), then I would inline a call to that:
            `virtual Command* clone() { return Command::clone(); }`
otherwise, the declaration in command.h should be pure virtual as well.

include/multikeycommand.h
```diff
@@ -0,0 +1,73 @@
...
+*          of the KeyedCommand class.
+*/
+class MultiKeyCommand : public KeyedCommand
+{
+    public:
+        /*!
+        *   \brief Run this Command on the RedisServer.
+        *   \param r A pointer to the RedisServer
+        */
+	    virtual CommandReply run_me(RedisServer* r);
```
billschereriii, line 52: single-letter variable names should only be for loop iterators. Name this something like `server`. (This comment applies to all the *command.[h,cpp] files.)

include/nonkeyedcommand.h
```diff
@@ -0,0 +1,143 @@
...
+        *            are copied using the copy assignment
+        *            operator for the derived type. This is meant
+        *            to provide functionality to deep
+        *            copy a Command.
+        *   \returns A pointer to dynamically allocated
+        *            derived type cast to parent Command
+        *            type.
+        */
+        virtual Command* clone() = 0;
+
```
billschereriii, line 96: Same comment as in keyedcommand

include/tensorbase.h
```diff
@@ -119,7 +119,7 @@ class TensorBase{
         *   \brief Deep copy operator
         *   \details This method creates a new derived
         *            type Tensor and returns a TensorBase*
-        *            pointer.  The new dervied type is
+        *            pointer.  The new derived type is
```
billschereriii, line 122: Awww, but dervied sounds so cute!

src/cpp/addressanycommand.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+CommandReply AddressAnyCommand::run_me(RedisServer* r)
+{
+    return r->run(*this);
+}
+
+Command* AddressAnyCommand::clone()
+{
+    AddressAnyCommand* new_cmd = new AddressAnyCommand(*this);
+    (*new_cmd) = *this;
```
billschereriii, line 42: Aren't these two lines redundant?

src/cpp/addressatcommand.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+CommandReply AddressAtCommand::run_me(RedisServer* r)
+{
+    return r->run(*this);
+}
+
+Command* AddressAtCommand::clone()
+{
+    AddressAtCommand* new_cmd = new AddressAtCommand(*this);
+    (*new_cmd) = *this;
```
billschereriii, line 42: Same -- I imagine this is repeated for each of the classes

include/clusterinfocommand.h
ctandon11: it is used when we declare the sub-class `class ClusterInfoCommand : public AddressAtCommand`

src/cpp/client.cpp
```diff
@@ -702,21 +696,41 @@ parsed_reply_map Client::get_db_cluster_info(std::string address)
...
+{
+    return this->_redis_server->run(cmd);
+}
+
+inline CommandReply Client::_run(CompoundCommand& cmd)
+{
+    return this->_redis_server->run(cmd);
+}
+
+inline std::vector<CommandReply> Client::_run(CommandList& cmds)
```
billschereriii, line 733: If these are inlines, then they should be implemented in the header, not in the cpp file. Either remove inline or move the implementation

include/keyedcommand.h
ctandon11: the declaration in command.h is pure virtual `virtual Command* clone() = 0;`

src/cpp/gettensorcommand.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "redisserver.h"
+
+using namespace SmartRedis;
+
+CommandReply GetTensorCommand::run_me(RedisServer* r)
+{
+    return r->run(*this);
+}
+
+std::vector<size_t> GetTensorCommand::get_tensor_dims(CommandReply& reply)
```
billschereriii, line 39: take "tensor" out of these method names -- it's redundant with the new class name

src/cpp/clusterinfocommand.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "clusterinfocommand.h"
+#include "redisserver.h"
+
+using namespace SmartRedis;
+
+parsed_reply_map ClusterInfoCommand::parse_db_cluster_info(std::string info)
+{
+    parsed_reply_map info_map;
+
+    std::string delim = "\r\n";
```
billschereriii, line 38: This is risky -- end-of-line varies depending on whether you're on UNIX (\r), Windows (\r\n), or Mac (\n). If we can run the server on different platforms, can we get different EOL sequences?

src/cpp/clusterinfocommand.cpp
```diff
@@ -0,0 +1,63 @@
...
+    std::string currKey = "";
```
billschereriii, line 39: If this variable isn't used, delete the declaration

src/cpp/compoundcommand.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+CommandReply CompoundCommand::run_me(RedisServer* r)
+{
+    return r->run(*this);
+}
+
+Command* CompoundCommand::clone()
+{
+    CompoundCommand* new_cmd = new CompoundCommand(*this);
+    (*new_cmd) = *this;
```
billschereriii, line 42: redundant

src/cpp/dbinfocommand.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "dbinfocommand.h"
+#include "redisserver.h"
+
+using namespace SmartRedis;
+
+parsed_reply_nested_map DBInfoCommand::parse_db_node_info(std::string info)
+{
+    parsed_reply_nested_map info_map;
+
+    std::string delim = "\r\n";
```
billschereriii, line 38: See previous comment. Probably just put a TODO in the code since I'm guessing you inherited this

src/cpp/gettensorcommand.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+    if(reply.n_elements() < 6)
```
billschereriii, line 41: It would be nice to have a named constant instead of 6 so a reader knows why we're looking for this many fields. Same with the 3s below.

src/cpp/gettensorcommand.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+    if(reply.n_elements() < 2)
+        throw std::runtime_error("The message does not have the correct "\
+                                "number of fields");
+
+    return TENSOR_TYPE_MAP.at(std::string(reply[1].str(),
+                                            reply[1].str_len()));
+}
+
+std::string_view GetTensorCommand::get_tensor_data_blob(CommandReply& reply)
```
billschereriii, line 65: General: Add a single-line comment before each method in the .cpp files that states what it does.  You can copy/paste the briefs from the headers for this.

src/cpp/nonkeyedcommand.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "nonkeyedcommand.h"
+#include "redisserver.h"
+
+using namespace SmartRedis;
+
+void NonKeyedCommand::set_exec_address_port(std::string address,
+                                    uint16_t port)
+{
+    this -> _address = address;
```
billschereriii, line 37: get rid of spaces around `->`

src/cpp/redis.cpp
```diff
@@ -49,62 +49,36 @@ Redis::~Redis()
...
-
-    while (n_trials > 0) {
+CommandReply Redis::run(SingleKeyCommand& cmd){
+    return this->_run(cmd);
+}
 
-        try {
-            reply = this->_redis->command(cmd_fields_start, cmd_fields_end);
+CommandReply Redis::run(MultiKeyCommand& cmd){
+    return this->_run(cmd);
```
billschereriii, line 57: If the implementation is so simple, why do we need a separate `_run()` method?

src/cpp/redis.cpp
```diff
@@ -144,7 +117,7 @@ CommandReply Redis::put_tensor(TensorBase& tensor)
 
 CommandReply Redis::get_tensor(const std::string& key)
 {
-    Command cmd;
+    GetTensorCommand cmd;
     cmd.add_field("AI.TENSORGET");
```
billschereriii, line 121: These could be moved to the GetTensorCommand constructor. Similar with other specialized commands

src/cpp/redis.cpp
```diff
@@ -320,13 +293,59 @@ CommandReply Redis::get_model(const std::string& key)
...
-    Command cmd;
+    SingleKeyCommand cmd;
     cmd.add_field("AI.SCRIPTGET");
     cmd.add_field(key, true);
     cmd.add_field("SOURCE");
     return this->run(cmd);
 }
 
+inline CommandReply Redis::_run(const Command& cmd)
+{
```
billschereriii, line 304: Hmm. We're getting the polymorphism with the different run overloads and then collapsing it back to a single `_run()` command. Is this actually buying us anything?

billschereriii: If not, not need to declare and implement the various overloads. It looks like it matters more for RedisCluster, so you can have the overloads there.

src/cpp/clusterinfocommand.cpp
ctandon11: The INFO [documentation](https://redis.io/commands/info) says "All the properties are in the form of field:value terminated by \r\n." Do you think that depends on the platform? I thought that implied its uniform across all platforms.

billschereriii: ok, cool. I still don't trust it completely, but it's good that it's separated out in a constant that we can generalize the code with if they ever change it out from under us

src/cpp/redis.cpp
ctandon11: the polymorphic functions are inherited from `RedisServer`, but are more useful for `RedisCluster` than `Redis`.  Right now, the `_run()` command is in `Redis` to avoid code repetition since all command types run the same way if they aren't being run on a cluster.

src/cpp/dbinfocommand.cpp
ctandon11: see comment above.

include/nonkeyedcommand.h
ctandon11: see comment above.

include/commandlist.tcc
```diff
@@ -0,0 +1,40 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SMARTREDIS_COMMANDLIST_TCC
+#define SMARTREDIS_COMMANDLIST_TCC
+
+template <class T>
+T* CommandList::add_command()
+{
+    T* new_cmd = new T();
```
billschereriii, line 35: Add:
```
if (new_cmd == NULL) {
    throw std::bad_alloc();
}
```

src/cpp/addressanycommand.cpp
```diff
@@ -0,0 +1,45 @@
...
+// Run AddressAnyCommand on the server
+CommandReply AddressAnyCommand::run_me(RedisServer* server)
+{
+    return server->run(*this);
+}
+
+// Deep copy operator
+Command* AddressAnyCommand::clone()
+{
+    AddressAnyCommand* new_cmd = new AddressAnyCommand(*this);
```
billschereriii, line 43: ...same for each instance where you're invoking operator new

src/cpp/nonkeyedcommand.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+using namespace SmartRedis;
+
+// Set address and port for command to be executed on
+void NonKeyedCommand::set_exec_address_port(std::string address,
+                                    uint16_t port)
+{
+    this->_address = address;
+    this->_port = port;
+    return;
```
billschereriii, line 40: Delete empty return statements at the end of routines

src/cpp/redis.cpp
```diff
@@ -49,62 +49,36 @@ Redis::~Redis()
...
-CommandReply Redis::run(CommandList& cmds)
+std::vector<CommandReply> Redis::run(CommandList& cmds)
 {
+    std::vector<CommandReply> replies;
     CommandList::iterator cmd = cmds.begin();
     CommandList::iterator cmd_end = cmds.end();
-    CommandReply reply;
     while(cmd != cmd_end) {
-        reply = this->run(**cmd);
+        replies.push_back(dynamic_cast<Command*>(*cmd)->run_me(this));
```
billschereriii, line 78: With this implementation, if any command fails badly enough to throw an exception, none of the subsequent ones will be attempted. I think this is what you want, but if you feel otherwise you should capture the result into a temporary `CommandReply` and put the call into a try/catch block.

billschereriii:
Library code update is approved for merge.  Unit test code needs review

include/commandlist.tcc
```diff
@@ -0,0 +1,43 @@
...
+    if (new_cmd == NULL){
```
mellis13, line 36: Doesn't the ``new`` operator throw ``bad_alloc`` unless ``std::no_throw`` is included in the ``new`` call?

billschereriii: This is theoretically true but I have never trusted compilers to be fully compliant on this point. I've seen really weird behavior from compilers in the past here. Given that we're writing code to work against a variety of compilers, this is a case where playing it safe is a good policy

include/keyedcommand.h
```diff
@@ -0,0 +1,108 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SMARTREDIS_KEYEDCOMMAND_H
+#define SMARTREDIS_KEYEDCOMMAND_H
+
+#include "command.h"
+#include "tensorbase.h"
+#include "enums/cpp_tensor_type.h"
```
mellis13, line 34: Do we need ``cpp_tensor_type.h`` in this class?

include/keyedcommand.h
```diff
@@ -0,0 +1,108 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SMARTREDIS_KEYEDCOMMAND_H
+#define SMARTREDIS_KEYEDCOMMAND_H
+
+#include "command.h"
+#include "tensorbase.h"
```
mellis13, line 33: Do we need ``tensorbase.h``

include/keyedcommand.h
```diff
@@ -0,0 +1,108 @@
...
+        *            derived type cast to parent Command
+        *            type.
+        */
+        virtual Command* clone() = 0;
+
+        /*!
+        *   \brief Run this Command on the RedisServer.
+        *   \param server A pointer to the RedisServer
+        */
+        virtual CommandReply run_me(RedisServer * server) = 0;
```
mellis13, line 103: ``*`` should go next to the type ``RedisServer``.

include/singlekeycommand.h
```diff
@@ -0,0 +1,75 @@
...
+*          of the KeyedCommand class.
+*/
+class SingleKeyCommand : public KeyedCommand
+{
+    public:
+        /*!
+        *   \brief Run this Command on the RedisServer.
+        *   \param server A pointer to the RedisServer
+        */
+	    virtual CommandReply run_me(RedisServer * server);
```
mellis13, line 54: Put ``*`` next to ``RedisServer``

src/cpp/client.cpp
```diff
@@ -682,12 +676,12 @@ parsed_reply_nested_map Client::get_db_node_info(std::string address)
...
         throw std::runtime_error(std::string(address) +
                                  "is not a valid database node address.");
-    Command cmd;
+    DBInfoCommand cmd;
     cmd.set_exec_address_port(host, port);
     cmd.add_field("INFO");
     cmd.add_field("everything");
     CommandReply reply = this->_run(cmd);
-    return CommandReplyParser::parse_db_node_info(std::string(reply.str(),
+    return cmd.parse_db_node_info(std::string(reply.str(),
```
mellis13, line 684: Since this is a static member function, should we call it like other static functions: ``DBInfoCommand::parse_db_node_info(reply)``?

src/cpp/client.cpp
```diff
@@ -702,25 +696,15 @@ parsed_reply_map Client::get_db_cluster_info(std::string address)
...
         throw std::runtime_error(std::string(address) +
                                  "is not a valid database node address.");
-    Command cmd;
+    ClusterInfoCommand cmd;
     cmd.set_exec_address_port(host, port);
     cmd.add_field("CLUSTER");
     cmd.add_field("INFO");
     CommandReply reply = this->_run(cmd);
-    return CommandReplyParser::parse_db_cluster_info(std::string(reply.str(),
+    return cmd.parse_db_cluster_info(std::string(reply.str(),
```
mellis13, line 704: Since ``parse_db_cluster_info`` is static should we call it like others: ``ClusterInfoCommand::parse_db_cluster_info(reply)``?

tests/cpp/unit-tests/test_addressanycommand.cpp
```diff
@@ -0,0 +1,79 @@
...
+                    REQUIRE(it_cpy != it_end_cpy);
+                    CHECK(*it == *it_cpy);
+                    it++;
+                    it_cpy++;
+                }
+                CHECK(it_cpy == it_end_cpy);
+
+                cmd_keys = cmd.get_keys();
+                std::vector<std::string> cmd_keys_cpy = cmd_cpy->get_keys();
+                std::sort(cmd_keys.begin(), cmd_keys.end());
```
mellis13, line 62: Out of curiosity, why are we sorting?  Is the order of ``Command`` keys not maintained in a copy operation?

tests/cpp/unit-tests/test_addressatcommand.cpp
```diff
@@ -0,0 +1,108 @@
...
+            char field_2[8] = "CLUSTER";
+            const char field_3[11] = "everything";
+            char field_4[5] = "INFO";
+            size_t field_size_4 = std::strlen(field_4);
+            std::string field_5 = "INFO";
+            std::vector<std::string> fields_1 = {"CLUSTER", "everything", "INFO"};
+            std::string_view field_sv = std::string_view(field_4, field_size_4);
+
+            // define necessary variables for testing
+            std::string output = " INFO CLUSTER everything INFO "
```
mellis13, line 52: Is the ``Command`` ``to_string()`` prepending with a space and that's why we need to format like this?

tests/cpp/unit-tests/test_addressatcommand.cpp
```diff
@@ -0,0 +1,108 @@
+#include "../../../third-party/catch/catch.hpp"
+#include "addressatcommand.h"
+
+using namespace SmartRedis;
+
+SCENARIO("Ensuring the iterators for an AddressAtCommand are correct", "[AddressAtCommand]")
```
mellis13, line 6: General comment: Should we also test member variables specific to ``AddressAtcommand`` and ``NonKeyedCommand`` (e.g. port number and host)?

tests/cpp/unit-tests/test_client.cpp
```diff
@@ -414,3 +423,34 @@ SCENARIO("Testing Client Object", "[Client]")
         }
     }
 }
+
+SCENARIO("Testing INFO Functions on Client Object", "[Client]")
+{
+
+    GIVEN("A Client object")
+    {
+        Client client(true);
```
mellis13, line 432: Should this client initialization use ``use_cluster()``?

tests/cpp/unit-tests/test_client.cpp
```diff
@@ -414,3 +423,34 @@ SCENARIO("Testing Client Object", "[Client]")
...
+                std::runtime_error);
+            CHECK_THROWS_AS(
+                client.get_db_cluster_info(db_address),
+                std::runtime_error);
+        }
+
+        AND_THEN("Calling INFO on database or cluster with "
+                "a valid address does not throw errors")
+        {
+            std::string db_address = "127.0.0.1:6379";
```
mellis13, line 450: This would only be true for a locally launched cluster / non-cluster.  I think the test suite has a method to read in the ``SSDB`` variable.

tests/cpp/unit-tests/test_client.cpp
```diff
@@ -414,3 +423,34 @@ SCENARIO("Testing Client Object", "[Client]")
...
+
+            CHECK_NOTHROW(client.get_db_node_info(db_address));
+            CHECK_NOTHROW(client.get_db_cluster_info(db_address));
```
mellis13, line 453: Won't this throw an error on a non-cluster testing environment?  Should we have an ``if`` statement here that uses use_cluster to check that behavior?

tests/cpp/unit-tests/test_multikeycommand.cpp
```diff
@@ -0,0 +1,51 @@
...
+            cmd.add_fields(fields_1, true);
+
+            THEN("The MultiKeyCommand object is structured correctly")
+            {
+                CHECK(cmd.has_keys() == true);
+                CHECK(cmd.first_field() == field_1);
+                CHECK(output == cmd.to_string());
+
+                cmd_keys = cmd.get_keys();
+                std::sort(cmd_keys.begin(), cmd_keys.end());
```
mellis13, line 46: Same comment as before on using sort.  If we choose not to sort, then we will need to change ``sorted_keys`` to align with this decision.

tests/cpp/unit-tests/test_singlekeycommand.cpp
```diff
@@ -0,0 +1,96 @@
...
+                REQUIRE(it_cpy != it_end_cpy);
+                CHECK(*it == *it_cpy);
+                it++;
+                it_cpy++;
+            }
+            CHECK(it_cpy == it_end_cpy);
+
+            cmd_keys = cmd->get_keys();
+            std::vector<std::string> cmd_keys_cpy = cmd_cpy->get_keys();
+            std::sort(cmd_keys.begin(), cmd_keys.end());
```
mellis13, line 79: Same comment as before on sorting keys.  Just flagging it again so we are consistent if we change it.

mellis13 (approved):
Looks good -- really great work!

PR #128, opened 2021-08-09
Add logic to create third-party directory if it doesn't exist

### LINKED ISSUE #112: Make test-deps fails after make clobber
# Description
make clobber deletes the third-party folder, but make test-deps assumes that it will be present

# How to reproduce
1. make clobber
2. make test-deps

# Expected behavior
Test dependencies should install and build

# System
any

billschereriii:
Fixes #112 by making the third-party if it doesn't exist yet


mellis13 (approved):
Good catch!  LGTM

Issue #59, opened 2021-06-07
CommandReply copy constructor and copy assignment operator


mellis13:
# Description
Currently the CommandReply copy constructor and copy assignment operator are not implemented.  However, as issue #1 is addressed, CommandReply should be fully implemented.  While passing CommandReply by reference is encouraged, copy operations should be enabled and allowed.

# Justification
This is a potential requirement for issue #1 as CommandReply objects are more broadly utilized for Python client commands.  Additionally, if CommandReply is ever exposed to the user for custom command execution, all operators should be implemented.

# Implementation Strategy

- [x] Implement copy assignment operator
- [x] Implement copy constructor

PR #138, opened 2021-08-20
Add copy constructors and copy assignment operators to CommandReply

EricGustin:
This PR addresses issue #59 along with some additional features.

Things added:

- CommandReply copy constructor. `CommandReply(const CommandReply& reply)`
- CommandReply copy constructor with `redisReply*` as input. `CommandReply(const redisReply* reply)`
- CommandReply copy assignment operator. `CommandReply& operator=(const CommandReply& reply)`
- CommandReply copy assignment operator with `redisReply*` as input. `CommandReply& operator=(const redisReply* reply)`
- Private helper function for deep copying a `redisReply*`.  `redisReply* deep_clone_reply(const redisReply* reply)`
- Unit tests that cover new features and slightly increase overall test coverage.

Things changed:

- The private CommandReply shallow copy constructor with `redisReply*` as input was renamed to `CommandReply shallow_clone(const redisReply* reply)`
- Made tab sizes consistent (size 4)

### DIFF
--- a/src/cpp/commandreply.cpp
+++ b/src/cpp/commandreply.cpp
@@ -30,10 +30,16 @@
 
 using namespace SmartRedis;
 
+CommandReply::CommandReply(const CommandReply& reply)
+{
+    _uptr_reply = RedisReplyUPtr(deep_clone_reply(reply._uptr_reply.get()), sw::redis::ReplyDeleter());
+    _reply = _uptr_reply.get();
+}
+
 CommandReply::CommandReply(redisReply* reply)
 {
-    this->_uptr_reply = 0;
-    this->_reply = reply;
+    _uptr_reply = RedisReplyUPtr(deep_clone_reply(reply), sw::redis::ReplyDeleter());
+    _reply = _uptr_reply.get();
 }
 
 CommandReply::CommandReply(RedisReplyUPtr&& reply)
@@ -57,6 +63,22 @@ CommandReply::CommandReply(CommandReply&& reply)
     return;
 }
 
+CommandReply& CommandReply::operator=(const CommandReply& reply)
+{
+    _uptr_reply.reset(NULL);
+    _uptr_reply = RedisReplyUPtr(deep_clone_reply(reply._uptr_reply.get()), sw::redis::ReplyDeleter());
+    _reply = _uptr_reply.get();
+    return *this;
+}
+
+CommandReply& CommandReply::operator=(redisReply* reply)
+{
+    _uptr_reply.reset(NULL);
+    _uptr_reply = RedisReplyUPtr(deep_clone_reply(reply), sw::redis::ReplyDeleter());
+    _reply = _uptr_reply.get();
+    return *this;
+}
+
 CommandReply& CommandReply::operator=(RedisReplyUPtr&& reply)
 {
     this->_uptr_reply = std::move(reply);
@@ -78,6 +100,14 @@ CommandReply& CommandReply::operator=(CommandReply&& reply)
         this->_reply = this->_uptr_reply.get();
     }
     return *this;
+}
+
+CommandReply CommandReply::shallow_clone(redisReply* reply) // shallow clone
+{
+    CommandReply r;
+    r._uptr_reply = 0;
+    r._reply = reply;
+    return r;
 }
 
 char* CommandReply::str()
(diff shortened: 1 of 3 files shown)


src/cpp/commandreply.cpp
```diff
@@ -30,236 +30,294 @@
...
+            }
+            break;
+        case REDIS_REPLY_ERROR:
+        case REDIS_REPLY_STATUS:
+        case REDIS_REPLY_STRING:
+        case REDIS_REPLY_DOUBLE:
+            // allocate memory for str and do deep copy
+            if (redis_reply->len > 0) {
+                redis_reply->str = new char[redis_reply->len];
+                std::strcpy(redis_reply->str, reply->str);
```
mellis13, line 316: I think we should do a memcpy here.  The string may contain null characters, and I think strcpy will terminate at the first null character.  We want to make sure we copy the whole string.

src/cpp/commandreply.cpp
```diff
@@ -30,236 +30,294 @@
...
+    redis_reply->str = NULL;
+    redis_reply->element = NULL;
+    switch (redis_reply->type) {
+        case REDIS_REPLY_ARRAY:
+        case REDIS_REPLY_MAP:
+        case REDIS_REPLY_SET:
+            // allocate memory for element and do deep copy
+            if (redis_reply->elements > 0) {
+                redis_reply->element = new redisReply*[redis_reply->elements];
+                if (reply->element != NULL) {
```
mellis13, line 303: This is a good catch to make sure we don't do ``reply->element[i]`` if ``reply`` is ``NULL``.  However, I wonder if we should have an else case?  

That is, if ``reply->element`` is NULL, that means ``reply->elements`` and ``reply->element`` are in inconsistent states.  If that is true, we will have ``redis_repy->element`` set to an array of ``redis_reply*`` structs with default values.  This would be inconsistent with the input ``reply`` object.  We should either free memory and set ``redis_reply->element`` to ``NULL`` or throw an error because the input reply is invalid.  What do you think?

src/cpp/commandreply.cpp
```diff
@@ -30,236 +30,294 @@
...
+    *redis_reply = *reply;
+    redis_reply->str = NULL;
+    redis_reply->element = NULL;
+    switch (redis_reply->type) {
+        case REDIS_REPLY_ARRAY:
+        case REDIS_REPLY_MAP:
+        case REDIS_REPLY_SET:
+            // allocate memory for element and do deep copy
+            if (redis_reply->elements > 0) {
+                redis_reply->element = new redisReply*[redis_reply->elements];
```
mellis13, line 302: Is it worth assigning a default value of ``NULL`` to the elements?  I don't think arrays assign default values during construction.

src/cpp/commandreply.cpp
```diff
@@ -30,236 +30,294 @@
...
-  this->_uptr_reply = std::move(reply._uptr_reply);
-  this->_reply = this->_uptr_reply.get();
-  return;
+    if (this != &reply) {
+        _uptr_reply = std::move(reply._uptr_reply);
+        _reply = _uptr_reply.get();
+    }
+}
+
+CommandReply& CommandReply::operator=(const CommandReply& reply)
```
mellis13, line 69: I might be wrong on this (maybe we can write a quick test), but for some of these``deep_clone_reply()`` calls is there a potential problem with scenarios of shallowed cloned ``CommandReply`` objects later being copied?  I sketched out a scenario below:

If I have ``CommandReply`` of type ``REDIS_REPLY_ARRAY `` returned from a redis++ call:
```c++
CommandReply first_reply = client.some_redis_call();
// first_ptr._uptr_reply = u_ptr_1
// first_ptr._reply = u_ptr_1.get()
```

Then if I call the ``operator[]`` function:

```c++
CommandReply sub_reply = first_reply[0];
//sub_reply._uptr_reply = NULL
//sub_reply._reply = u_ptr_1.get()
```

Then if I call the copy constructor:
```
CommandReply copy_sub_reply(sub_reply);
```
Are ``_uptr_reply`` and ``_reply`` of ``copy_sub_reply`` both ``NULL``?

src/cpp/commandreply.cpp
mellis13: I actually tested this, and it looks like it is OK.  Resolving this conversation.

Issue #122, opened 2021-08-07
Expose CONFIG SET to the python client


ctandon11:
# Description
Give the python client access to the Redis command [CONFIG SET](https://redis.io/commands/config-set) by exposing it
through the Pybindings.

# Justification
The user will be able to reconfigure the server at run time without
needing to restart Redis.

# Implementation Strategy
The C++ client has this functionality so just needs to be exposed through the Pybindings
and `pyclient.cpp` file to the python client layer. This will be an `AddressAtCommand`.

Issue #121, opened 2021-08-07
Expose CONFIG GET to the python client


ctandon11:
# Description
Give the python client access to the Redis command [CONFIG GET](https://redis.io/commands/config-get) by exposing it
through the Pybindings.

# Justification
The user will be able to access the configuration parameters of a running
Redis Server.

# Implementation Strategy
The C++ client has this functionality so just needs to be exposed through the Pybindings
and `pyclient.cpp` file to the python client layer. This will be an `AddressAtCommand`. A
function will likely be needed to parse through the `CommandReply`.

Issue #117, opened 2021-08-06
Expose FLUSHDB to the python client


ctandon11:
# Description
Give the python client access to the Redis command [FLUSHDB](https://redis.io/commands/flushdb) by exposing it
through the Pybindings.

# Justification
Users will be able delete all the keys of the given databases from the
python client.

# Implementation Strategy
The C++ client has this functionality so just needs to be exposed through the Pybindings
and `pyclient.cpp` file to the python client layer. This will be an `AddressAtCommand`
in which the user will input the address of a db node.

Issue #16, opened 2021-03-17
Custom error handling


mellis13:
**Description**
Currently we throw ```std::runtime``` errors for all internal C++ client errors, including those that result for redis-plus-plus and hiredis errors.  This was done initially for convenience so that we could easily handle them in the Python client.  However, as client functionality grows, we need better visibility for users and developers into the errors being thrown so that better exception handling can be performed.  

**Implementation Strategy**

Custom error classes should be implemented in core C++ client, and those errors should be wrapped into the Python client binding.

PR #139, opened 2021-08-25
Expose FLUSHDB, CONFIG GET, and CONFIG SET to the python client

ctandon11:
This PR exposes the Redis command [FLUSHDB](https://redis.io/commands/flushdb), 
[CONFIG GET](https://redis.io/commands/config-get), and [CONFIG SET](https://redis.io/commands/config-set) to the python client. 
It addresses issues #117 , #122, and #121 which are part of epic #115.

`FLUSHDB` is an `AddressAtCommand` that returns `"OK"` when
it is executed properly. In the python client, the user can call `FLUSHDB` 
on a list of databases. Unit tests are added to ensure that all the keys 
from the given database(s) are deleted when `FLUSHDB` is called.

`CONFIG SET` is an `AddressAtCommand` that returns `"OK"` when
it is executed properly. Tests are added to ensure `CONFIG SET`
is called on a valid address and that it actually sets configuration parameters.

`CONFIG GET` is an `AddressAtCommand` that returns a vector of pairs containing 
configuration parameters and their values (i.e. [(config_param1: value1), (config_param2, value2), ...] . 
Tests are added to ensure `CONFIG SET` is called on a valid address.

### DIFF
--- a/tests/cpp/unit-tests/test_client.cpp
+++ b/tests/cpp/unit-tests/test_client.cpp
@@ -548,7 +548,7 @@ SCENARIO("Testing FLUSHDB on Client Object", "[Client][FLUSHDB]")
             THEN("The database is flushed")
             {
                 // ensure the database has things to flush
-                CHECK_FALSE(client.dataset_exists(dataset_name) == true);
+                CHECK(client.dataset_exists(dataset_name) == true);
                 CHECK(client.tensor_exists(tensor_key) == true);
                 // flush the database
                 std::string db_address = parse_SSDB(std::getenv("SSDB"));


include/addressatcommand.h
```diff
@@ -66,6 +66,28 @@ class AddressAtCommand : public NonKeyedCommand
...
         virtual Command* clone();
+
+        /*!
+        *   \brief Returns host of database node
+        *   \param address The address of the database node
+        *   \returns The host of the database node
+        */
+        inline std::string parse_host(std::string address)
+        {
+            return address.substr(0, address.find(":"));
```
mellis13, line 77: It looks like ``substr`` can return ``out_of_range`` and ``bad_alloc`` exceptions.  Maybe we should account for those here (and below in parse_host) and either rethrow as is or as ``std::runtime_error`` so that when #16 is complete better handle failure with new custom error objects?  Also, do you think it is worth checking to make sure that ":" is in the string and not the first position (and if not) throw an error?

include/addressatcommand.h
```diff
@@ -66,6 +66,28 @@ class AddressAtCommand : public NonKeyedCommand
...
+        }
+
+        /*!
+        *   \brief Returns port of database node
+        *   \param address The address of the database node
+        *   \returns The port of the database node
+        */
+        inline uint64_t parse_port(std::string address)
+        {
+            return std::stoul(address.substr(address.find(":") + 1),
```
mellis13, line 87: Similar comment as ``parse_host`` except the additional check suggestion would be to make sure ``:`` exists and isn't the last position so we don't return empty string.

include/client.h
```diff
@@ -567,19 +567,53 @@ class Client
...
+        *   \returns "OK" if keys are successfully flushed
+        */
+        std::string flush_db(std::string address);
+
+        /*!
+        *   \brief Read the configuration parameters of a running server.
+        *   \param expression Parameter used in the configuration or a
+        *                     glob pattern (Use '*' to retrieve all
+        *                     configuration parameters)
+        *   \param address The address of the database node execute on
```
mellis13, line 594: Change ``execute` on`` to ``to execute on``?

include/client.h
```diff
@@ -567,19 +567,53 @@ class Client
...
+
+        /*!
+        *   \brief Reconfigure the server. It can change both trivial
+        *          parameters or switch from one to another persistence option.
+        *          All the configuration parameters set using this command are
+        *          immediately loaded by Redis and will take effect starting with
+        *          the next command executed.
+        *   \param config_param A configuration parameter to set
+        *   \param value The value to assign to the configuration parameter
+        *   \param address The address of the database node execute on
```
mellis13, line 610: Change ``execute on`` to ``to execute on``?

Issue #103, opened 2021-07-29
C++ Code Cleanup


billschereriii:
**Description**

Assorted code cleanup:
1.  Put try/catch blocks around C++ code, mainly around new operator and memory allocation in constructors
2. Null pointer checks and pointer validation
3. Initialize all variables; do not pass uninitialized variables to other routines
* Define default values for enums so that they can be passed in as initialized 
4. Use copy constructors for all STL containers
* A lot of code uses push_back instead of copy constructors
values into C functions (e.g. line 121 of get_dataset_tensor()
5. Use NULL, not 0, for pointers
6. NULL out pointers after deallocation
7. Move error messages to a named constants in a separate header in support of eventual internationalization
8. Switch from C-style casts to C++ casts
9. Make sure that const usage is correct and consistent
10. Make sure that error conditions are handled properly and locally
11. Make sure memory management strategy is fully document for callers
12. rename variables prefixed "g_" to avoid confusion with Hungarian notation (in which this would mean the variables are globals)
13. Refactor code in Client::poll_key(), Client::poll_model(), Client::poll_tensor()
* Investigate use of busy-waiting for poll methods for small intervals
14. Make an inline for prefix management in Client code
15. Unify memory management to all malloc/free or new/delete
16. Eliminate unnecessary Command::add_field() overload
17. Use standard iterators in Command::add_fields()
18. Replace while loops with for loops where it improves readability
19. Refactor duplicate code into MetaData::Clone()
20. Refactor MetaData::add_scalar() for efficiency
21. Check for type mismatch in MetaData::add_string()
22. Add default cases for all switch statements
23. Make metadatafield a struct instead of a class; make members public
24. Investigate use of environment variables to add user control over Redis::run() behavior. 
25. Make sure compiler warnings are at the highest finickiness
26. RedisServer::_get_ssdb(): Consider supporting round-robin allocation

PR #141, opened 2021-08-26
Cpp cleanup

billschereriii:
Addresses Issue #103.

### DIFF
--- a/include/enums/c_tensor_type.h
+++ b/include/enums/c_tensor_type.h
@@ -33,7 +33,7 @@
 #define SMARTREDIS_CTENSORTYPE_H
 
 typedef enum {
-	c_invalid_tensor = 0,
+    c_invalid_tensor = 0,
     c_dbl     = 1,
     c_flt     = 2,
     c_int8    = 3,
--- a/include/enums/cpp_memory_layout.h
+++ b/include/enums/cpp_memory_layout.h
@@ -59,6 +59,8 @@ inline MemoryLayout convert_layout(CMemoryLayout layout) {
       return MemoryLayout::contiguous;
     case c_fortran_contiguous:
       return MemoryLayout::fortran_contiguous;
+    case c_invalid_layout:
+      return MemoryLayout::invalid;
     default:
       throw std::runtime_error("Unsupported enum conversion.");
   }
--- a/include/enums/cpp_metadata_type.h
+++ b/include/enums/cpp_metadata_type.h
@@ -47,66 +47,52 @@ enum class MetaDataType {
 
 //! Helper method to convert between MetaDataType and CMetaDataType
 inline CMetaDataType convert_metadata_type(MetaDataType type) {
-    CMetaDataType t;
     switch (type) {
+        case MetaDataType::invalid:
+            return CMetaDataType::c_meta_invalid;
         case MetaDataType::dbl:
-            t = CMetaDataType::c_meta_dbl;
-            break;
+            return CMetaDataType::c_meta_dbl;
         case MetaDataType::flt:
-            t = CMetaDataType::c_meta_flt;
-            break;
+            return CMetaDataType::c_meta_flt;
         case MetaDataType::int32:
-            t = CMetaDataType::c_meta_int32;
-            break;
+            return CMetaDataType::c_meta_int32;
         case MetaDataType::int64:
-            t = CMetaDataType::c_meta_int64;
-            break;
+            return CMetaDataType::c_meta_int64;
         case MetaDataType::uint32:
-            t = CMetaDataType::c_meta_uint32;
-            break;
+            return CMetaDataType::c_meta_uint32;
         case MetaDataType::uint64:
-            t = CMetaDataType::c_meta_uint64;
-            break;
+            return CMetaDataType::c_meta_uint64;
         case MetaDataType::string:
-            t = CMetaDataType::c_meta_string;
-            break;
+            return CMetaDataType::c_meta_string;
         default :
             throw std::runtime_error("Error converting MetaDataType "\
                                      "to CMetaDataType.");
     }
-    return t;
 }
 
 //! Helper method to convert between CMetaDataType and MetaDataType
 inline MetaDataType convert_metadata_type(CMetaDataType type) {
-    MetaDataType t;
     switch (type) {
+        case CMetaDataType::c_meta_invalid:
+            return MetaDataType::invalid;
         case CMetaDataType::c_meta_dbl:
-            t = MetaDataType::dbl;
-            break;
+            return MetaDataType::dbl;
         case CMetaDataType::c_meta_flt:
-            t = MetaDataType::flt;
-            break;
+            return MetaDataType::flt;
         case CMetaDataType::c_meta_int32:
-            t = MetaDataType::int32;
-            break;
+            return MetaDataType::int32;
         case CMetaDataType::c_meta_int64:
-            t = MetaDataType::int64;
-            break;
+            return MetaDataType::int64;
         case CMetaDataType::c_meta_uint32:
-            t = MetaDataType::uint32;
-            break;
+            return MetaDataType::uint32;
         case CMetaDataType::c_meta_uint64:
-            t = MetaDataType::uint64;
-            break;
+            return MetaDataType::uint64;
         case CMetaDataType::c_meta_string:
-            t = MetaDataType::string;
-            break;
+            return MetaDataType::string;
         default :
             throw std::runtime_error("Error converting CMetaDataType "\
                                      "to MetaDataType.");
     }
-    return t;
 }
 
 } //namespace SmartRedis
--- a/include/enums/cpp_tensor_type.h
+++ b/include/enums/cpp_tensor_type.h
@@ -51,73 +51,56 @@ enum class TensorType{
 
 //! Helper method to convert between CTensorType and TensorType
 inline TensorType convert_tensor_type(CTensorType type) {
-    TensorType t;
-    switch (type) {
+    switch (type) {             
+        case CTensorType::c_invalid_tensor:
+            return TensorType::undefined;
         case CTensorType::c_dbl:
-            t = TensorType::dbl;
-            break;
+            return TensorType::dbl;
         case CTensorType::c_flt:
-            t = TensorType::flt;
-            break;
+            return TensorType::flt;
         case CTensorType::c_int8:
-            t = TensorType::int8;
-            break;
+            return TensorType::int8;
         case CTensorType::c_int16:
-            t = TensorType::int16;
-            break;
+            return TensorType::int16;
         case CTensorType::c_int32:
-            t = TensorType::int32;
-            break;
+            return TensorType::int32;
         case CTensorType::c_int64:
-            t = TensorType::int64;
-            break;
+            return TensorType::int64;
         case CTensorType::c_uint8:
-            t = TensorType::uint8;
-            break;
+            return TensorType::uint8;
         case CTensorType::c_uint16:
-            t = TensorType::uint16;
-            break;
+            return TensorType::uint16;
         default:
             throw std::runtime_error("Error converting CTensorType "\
                                      "to TensorType.");
     }
-    return t;
 }
 
 //! Helper method to convert between CTensorType and TensorType
 inline CTensorType convert_tensor_type(TensorType type) {
-    CTensorType t;
     switch (type) {
+        case TensorType::undefined:
+            return CTensorType::c_invalid_tensor;
         case TensorType::dbl:
-            t = CTensorType::c_dbl;
-            break;
+            return CTensorType::c_dbl;
         case TensorType::flt:
-            t = CTensorType::c_flt;
-            break;
+            return CTensorType::c_flt;
         case TensorType::int8:
-            t = CTensorType::c_int8;
-            break;
+            return CTensorType::c_int8;
         case TensorType::int16:
-            t = CTensorType::c_int16;
-            break;
+            return CTensorType::c_int16;
         case TensorType::int32:
-            t = CTensorType::c_int32;
-            break;
+            return CTensorType::c_int32;
         case TensorType::int64:
-            t = CTensorType::c_int64;
-            break;
+            return CTensorType::c_int64;
         case TensorType::uint8:
-            t = CTensorType::c_uint8;
-            break;
+            return CTensorType::c_uint8;
         case TensorType::uint16:
-            t = CTensorType::c_uint16;
-            break;
+            return CTensorType::c_uint16;
         default :
             throw std::runtime_error("Error converting TensorType "\
                                      "to CTensorType.");
     }
-    return t;
-
 }
 
 } //namespace SmartRedis
--- a/include/scalarfield.tcc
+++ b/include/scalarfield.tcc
@@ -31,8 +31,8 @@
 
 // ScalarField constructor
 template <class T>
-ScalarField<T>::ScalarField(const std::string& name, MetaDataType type) :
-    MetadataField(name, type)
+ScalarField<T>::ScalarField(const std::string& name, MetaDataType type)
+  : MetadataField(name, type)
 {
     // NOP
 }
--- a/include/tensor.tcc
+++ b/include/tensor.tcc
@@ -74,7 +74,7 @@ Tensor<T>& Tensor<T>::operator=(const Tensor<T>& tensor)
     // Deep copy tensor data
     TensorBase::operator=(tensor);
     _set_tensor_data(tensor._data, tensor._dims,
-                            MemoryLayout::contiguous);
+                     MemoryLayout::contiguous);
     _c_mem_views = tensor._c_mem_views;
     _f_mem_views = tensor._f_mem_views;
 
@@ -141,15 +141,14 @@ void* Tensor<T>::data_view(const MemoryLayout mem_layout)
             ptr = _data;
             break;
         case MemoryLayout::fortran_contiguous:
-            ptr = _f_mem_views.allocate_bytes(
-                                     _n_data_bytes());
+            ptr = _f_mem_views.allocate_bytes(_n_data_bytes());
             _c_to_f_memcpy((T*)ptr, (T*)_data, _dims);
             break;
         case MemoryLayout::nested:
             _build_nested_memory(&ptr,
-                                       _dims.data(),
-                                       _dims.size(),
-                                       (T*)_data);
+                                 _dims.data(),
+                                 _dims.size(),
+                                 (T*)_data);
             break;
         default:
             throw std::runtime_error(
@@ -185,7 +184,6 @@ void Tensor<T>::fill_mem_space(void* data,
     }
 
     // Make sure there is space for all the data
-    // ***WS*** TBD: we should be checking each dimension here as well
     if (n_values != num_values()) {
         throw std::runtime_error("The provided dimensions do "\
                                   "not match the size of the "\
@@ -227,7 +225,7 @@ void* Tensor<T>::_copy_nested_to_contiguous(void* src_data,
         for (size_t i = 0; i < dims[0]; i++) {
           dest_data =
             _copy_nested_to_contiguous(*current, &dims[1],
-                                             n_dims-1, dest_data);
+                                       n_dims-1, dest_data);
           current++;
         }
     }
--- a/src/cpp/dataset.cpp
+++ b/src/cpp/dataset.cpp
@@ -45,8 +45,7 @@ void DataSet::add_tensor(const std::string& name,
                          const TensorType type,
                          MemoryLayout mem_layout)
 {
-    _add_to_tensorpack(name, data, dims,
-                             type, mem_layout);
+    _add_to_tensorpack(name, data, dims, type, mem_layout);
     _metadata.add_string(".tensor_names", name);
 }
 


CMakeLists.txt
```diff
@@ -6,8 +6,9 @@ option(BUILD_PYTHON "Build the python module" ON)
 
 set(CMAKE_BUILD_TYPE RELEASE)
 set(CMAKE_CXX_STANDARD 17)
-set(CMAKE_EXE_LINKER_FLAGS "-lpthread")
 set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR}/install)
+add_link_options(-lpthread)
+add_compile_options(-Wall -Werror)
```
mellis13, line 11: If we are going to turn on ``-Wall``, we should probably include a list of gcc versions and cray pe compiler versions in this PR that we have tested the build on to make sure users don't start getting build failures when we push to ``developer``.

include/enums/c_tensor_type.h
```diff
@@ -33,15 +33,15 @@
...
-    c_dbl       = 1,
-    c_flt       = 2,
-    c_int8      = 3,
-    c_int16     = 4,
-    c_int32     = 5,
-    c_int64     = 6,
-    c_uint8     = 7,
-    c_uint16    = 8
+	c_invalid_tensor = 0,
+    c_dbl     = 1,
```
mellis13, line 37: Align ``c_cbl`` and other enums with ``c_invalid_tensor``.  I think it is just this group misaligned.  Previous c enum declarations look good.

include/enums/cpp_memory_layout.h
```diff
@@ -51,19 +52,15 @@ inline MemoryLayout convert_layout(CMemoryLayout layout) {
   direct use of MemoryLayout.
   */
 
-  switch(layout) {
-    case(c_nested) :
+  switch (layout) {
```
mellis13, line 55: We should probably add ``invalid`` to the enum conversion switch cases.

include/enums/cpp_metadata_type.h
```diff
@@ -47,26 +48,26 @@ enum class MetaDataType {
 //! Helper method to convert between MetaDataType and CMetaDataType
 inline CMetaDataType convert_metadata_type(MetaDataType type) {
     CMetaDataType t;
-    switch(type) {
-        case MetaDataType::dbl :
+    switch (type) {
```
mellis13, line 51: We should probably add the ``invalid`` type to these switch cases too.

include/enums/cpp_metadata_type.h
```diff
@@ -79,26 +80,26 @@ inline CMetaDataType convert_metadata_type(MetaDataType type) {
 //! Helper method to convert between CMetaDataType and MetaDataType
 inline MetaDataType convert_metadata_type(CMetaDataType type) {
     MetaDataType t;
-    switch(type) {
-        case CMetaDataType::c_meta_dbl :
+    switch (type) {
```
mellis13, line 83: We should probably add the ``invalid`` type to these switch cases too.

include/enums/cpp_tensor_type.h
```diff
@@ -52,32 +52,32 @@ enum class TensorType{
 //! Helper method to convert between CTensorType and TensorType
 inline TensorType convert_tensor_type(CTensorType type) {
     TensorType t;
-    switch(type) {
-        case CTensorType::c_dbl :
+    switch (type) {
```
mellis13, line 55: We should probably add the ``invalid`` type to these switch cases too.  Also, maybe the ``0`` value for TensorType should be ``invalid`` to match the naming convention used in other enums?

include/enums/cpp_tensor_type.h
```diff
@@ -87,29 +87,29 @@ inline TensorType convert_tensor_type(CTensorType type) {
 //! Helper method to convert between CTensorType and TensorType
 inline CTensorType convert_tensor_type(TensorType type) {
     CTensorType t;
-    switch(type) {
-        case TensorType::dbl :
+    switch (type) {
```
mellis13, line 90: We should probably add the ``invalid`` type to these switch cases too.

src/cpp/addressanycommand.cpp
```diff
@@ -41,8 +41,10 @@ CommandReply AddressAnyCommand::run_me(RedisServer* server)
...
-    if (new_cmd == NULL){
+    if (new_cmd == NULL) {
         throw std::bad_alloc();
     }
     return new_cmd;
-}
\ No newline at end of file
+}
+
+// EOF
```
mellis13, line 50: I might skip the ``// EOF `` comment since we aren't doing it in other files.

src/cpp/addressatcommand.cpp
```diff
@@ -41,8 +41,10 @@ CommandReply AddressAtCommand::run_me(RedisServer* server)
...
 {
     AddressAtCommand* new_cmd = new AddressAtCommand(*this);
-    if (new_cmd == NULL){
+    if (new_cmd == NULL) {
         throw std::bad_alloc();
     }
     return new_cmd;
 }
+
+// EOF
```
mellis13, line 50: I might skip the ``// EOF `` comment since we aren't doing it in other files.

CMakeLists.txt
billschereriii: Agree, though this is independently true whether we have this PR or not

include/enums/c_memory_layout.h
```diff
@@ -34,11 +34,12 @@ as a specifier.
 #ifndef SMARTREDIS_CMEMORYLAYOUT_H
 #define SMARTREDIS_CMEMORYLAYOUT_H
 
-typedef enum{
-    c_nested=1,
-    c_contiguous=2,
-    c_fortran_nested=3,
-    c_fortran_contiguous=4
-}CMemoryLayout;
+typedef enum {
```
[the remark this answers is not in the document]
billschereriii: Shouldn't be too hard. I'll tackle that as part of the error handling PR

include/enums/cpp_memory_layout.h
billschereriii: I'm surprised I didn't get a warning (and thus fail to compile) with this

src/cpp/addressanycommand.cpp
billschereriii: I think I added it to all the files

src/cpp/addressatcommand.cpp
billschereriii: see above

include/scalarfield.tcc
```diff
@@ -29,71 +29,76 @@
 #ifndef SMARTREDIS_SCALARFIELD_TCC
 #define SMARTREDIS_SCALARFIELD_TCC
 
+// ScalarField constructor
 template <class T>
-ScalarField<T>::ScalarField(const std::string& name,
-                            MetaDataType type) :
+ScalarField<T>::ScalarField(const std::string& name, MetaDataType type) :
```
mellis13, line 34: Not a big deal, but it looks like in other functions below you put the ``:`` on the next line.  Either is fine, probably just better to be consistent

include/sharedmemorylist.tcc
```diff
@@ -29,27 +29,29 @@
...
+    _inventory.push_front(s_ptr);
 }
 
+// Allocate memory and record the allocation
 template <class T>
 T* SharedMemoryList<T>::allocate_bytes(size_t bytes)
 {
-    T* ptr = (T*)malloc(bytes);
-    this->add_allocation(bytes, ptr);
+    T* ptr = (T*)new unsigned char[bytes];
```
mellis13, line 44: Is there a functional difference between ``char`` and ``unsigned char`` for byte allocation?  If not functional difference, just using ``char`` might be better.

include/tensor.tcc
```diff
@@ -37,57 +38,79 @@ Tensor<T>::Tensor(const std::string& name,
...
+    // Done
     return *this;
 }
 
+// Deep copy operator
 template <class T>
 TensorBase* Tensor<T>::clone()
 {
     Tensor<T>* new_tensor = new Tensor<T>(*this);
+    if (new_tensor == NULL)
```
mellis13, line 107: Isn't the standard that ``new`` will throw ``bad_alloc`` unless ``nothrow`` is passed in?  I think programming to non-standard behavior is going to be a pain to enforce and inconsistent.  If there is an example of a compiler not doing standard behavior, we should put this is a wrapper function or something that is easier to maintain and enforce.

include/tensor.tcc
```diff
@@ -113,140 +136,154 @@ void* Tensor<T>::data_view(const MemoryLayout mem_layout)
...
+    switch (mem_layout) {
+        case MemoryLayout::contiguous:
+            ptr = _data;
             break;
-        case(MemoryLayout::fortran_contiguous) :
-            ptr = this->_f_mem_views.allocate_bytes(
-                                     this->_n_data_bytes());
-            this->_c_to_f_memcpy((T*)ptr, (T*)this->_data, this->_dims);
+        case MemoryLayout::fortran_contiguous:
+            ptr = _f_mem_views.allocate_bytes(
```
mellis13, line 146: Probably fits on a single line of text without ``this``.

include/tensor.tcc
```diff
@@ -113,140 +136,154 @@ void* Tensor<T>::data_view(const MemoryLayout mem_layout)
...
+            _c_to_f_memcpy((T*)ptr, (T*)_data, _dims);
             break;
-        case(MemoryLayout::nested) :
-            this->_build_nested_memory(&ptr,
-                                       this->_dims.data(),
-                                       this->_dims.size(),
-                                       (T*)this->_data);
+        case MemoryLayout::nested:
+            _build_nested_memory(&ptr,
+                                       _dims.data(),
```
mellis13, line 152: Align text

include/tensor.tcc
```diff
@@ -37,57 +38,79 @@ Tensor<T>::Tensor(const std::string& name,
...
-        this->_f_mem_views = tensor._f_mem_views;
-    }
+    // Check for self assignment
+    if (this == &tensor)
+        return *this;
+
+    // Deep copy tensor data
+    TensorBase::operator=(tensor);
+    _set_tensor_data(tensor._data, tensor._dims,
+                            MemoryLayout::contiguous);
```
mellis13, line 77: Align text.

include/tensor.tcc
```diff
@@ -113,140 +136,154 @@ void* Tensor<T>::data_view(const MemoryLayout mem_layout)
...
+    for ( ; it != dims.cend(); it++) {
+        if (*it <= 0) {
+            throw std::runtime_error("All dimensions must be greater than 0.");
+        }
+        n_values *= (*it);
     }
 
-    if(n_values!=this->num_values())
+    // Make sure there is space for all the data
+    // ***WS*** TBD: we should be checking each dimension here as well
```
mellis13, line 190: We should resolve this comment before merging.

include/tensor.tcc
```diff
@@ -113,140 +136,154 @@ void* Tensor<T>::data_view(const MemoryLayout mem_layout)
...
                                             void* dest_data)
 {
-    if(n_dims > 1) {
+    if (n_dims > 1) {
         T** current = (T**)src_data;
-        for(size_t i = 0; i < dims[0]; i++) {
+        for (size_t i = 0; i < dims[0]; i++) {
           dest_data =
-            this->_copy_nested_to_contiguous(*current, &dims[1],
+            _copy_nested_to_contiguous(*current, &dims[1],
```
mellis13, line 231: Align text

include/tensor.tcc
```diff
@@ -113,140 +136,154 @@ void* Tensor<T>::data_view(const MemoryLayout mem_layout)
...
-                this->_build_nested_memory((void**)(&new_data[i]),
-                                           &dims[1], n_dims-1,
-                                           contiguous_mem);
+    if (dims == NULL) {
+        throw std::runtime_error("Missing dims in call to "\
+                                 "_build_nested_memory");
+    }
+    if (n_dims > 1) {
+        T** new_data = _c_mem_views.allocate(dims[0]);
+        if (new_data == NULL)
```
mellis13, line 280: Same comment as before on non-standard check.  But also, this check seems like it should go inside of ``SharedMemoryList`` near the ``new`` command and not here.

include/sharedmemorylist.tcc
billschereriii: unsigned char is the C/C++ standard for bytes

include/scalarfield.tcc
billschereriii: I tried to get them all moved to the next line to match the style recommendation from the Google style docs. Thanks for pointing out cases where I missed it

include/tensor.tcc
billschereriii: It's really hard for me to let go of this one because of how badly I was burned by it, but to be fair the failure I saw was over a decade ago, so it's probably time for me to move on...

include/tensor.tcc
billschereriii: ok

include/tensor.tcc
billschereriii: agree

include/tensor.tcc
billschereriii: This case is different -- we don't know what failure conditions _c_mem_views.allocate() has, so the NULL check is actually just validating the return value. Basic sanity checking on all returned pointers is good defensive practice

src/cpp/command.cpp
```diff
@@ -30,56 +30,61 @@
...
+    for (; ptr_it != _ptr_fields.end(); ptr_it++)
+        _fields[ptr_it->second] = ptr_it->first;
+
+    // copy _local_fields and _cmd_keys, and put the fields into _fields
+    std::vector<std::pair<char*, size_t> >::const_iterator local_it =
+        cmd._local_fields.begin();
+    for (; local_it != cmd._local_fields.end(); local_it++) {
+        // allocate memory and copy a local field
+        size_t field_size = cmd._fields[local_it->second].size();
+        char* f = new char[field_size];
```
mellis13, line 63: If we are going with ``unsigned char`` over ``char``, update this

src/cpp/command.cpp
```diff
@@ -30,56 +30,61 @@
...
+        if (f == NULL)
```
mellis13, line 64: Same comment as before.

src/cpp/command.cpp
```diff
@@ -90,43 +95,21 @@ void Command::add_field(std::string field, bool is_key)
     keys.
     */
 
-    int field_size = field.size();
-    char* f = (char*) malloc(sizeof(char)*(field_size+1));
+    size_t field_size = field.size();
+    char* f = (char*)new unsigned char[field_size + 1];
+    if (f == NULL)
```
mellis13, line 100: Same comment regarding non-standard check

src/cpp/command.cpp
```diff
@@ -137,19 +120,20 @@ void Command::add_field(const char* field, bool is_key)
...
     */
 
-    int field_size = std::strlen(field);
-    char* f = (char*) malloc(sizeof(char)*(field_size));
-    std::memcpy(f, field, sizeof(char)*field_size);
-    this->_local_fields.push_back({f, _fields.size()});
-    this->_fields.push_back(std::string_view(f, field_size));
+    size_t field_size = std::strlen(field);
+    char* f = new char[field_size];
+    if (f == NULL)
```
mellis13, line 125: This one too.  I'll probably stop tagging these and we just search for this syntax if we want to chage.

src/cpp/commandlist.cpp
```diff
@@ -30,63 +30,76 @@
...
 CommandList::CommandList(const CommandList& cmd_lst)
 {
     std::vector<Command*>::const_iterator c_it = cmd_lst._commands.cbegin();
-    std::vector<Command*>::const_iterator c_it_end = cmd_lst._commands.cend();
-    Command* new_cmd;
-    for(; c_it != c_it_end; c_it++) {
-        new_cmd = (*c_it)->clone();
+    for ( ; c_it != cmd_lst._commands.cend(); c_it++) {
+        Command* new_cmd = (*c_it)->clone();
+        if (new_cmd == NULL) {
```
mellis13, line 39: I don't think we need this check.  It doesn't seem like the ``clone()`` functions could ever return a NULL pointer (and we should enforce that they don't because we have full control of the API).

src/cpp/commandlist.cpp
```diff
@@ -30,63 +30,76 @@
...
-        Command* new_cmd;
-        while(c_it != c_it_end) {
-            new_cmd = (*c_it)->clone();
-            this->_commands.push_back(new_cmd);
-            c_it++;
+    // Copy over the new contents
+    std::vector<Command*>::const_iterator c_it = cmd_lst._commands.begin();
+    for ( ; c_it != cmd_lst._commands.end(); c_it++) {
+        Command* new_cmd = (*c_it)->clone();
+        if (new_cmd == NULL) {
```
mellis13, line 63: Same comment as before.

src/cpp/dataset.cpp
```diff
@@ -31,54 +31,68 @@
...
 
+// Add a tensor to the DataSet.
 void DataSet::add_tensor(const std::string& name,
                          void* data,
                          const std::vector<size_t>& dims,
                          const TensorType type,
                          MemoryLayout mem_layout)
 {
-    this->_add_to_tensorpack(name, data, dims,
+    _add_to_tensorpack(name, data, dims,
```
mellis13, line 48: Align text.

src/cpp/dataset.cpp
```diff
@@ -31,54 +31,68 @@
...
                          TensorType& type,
                          MemoryLayout mem_layout)
 {
-    this->_enforce_tensor_exists(name);
+    _enforce_tensor_exists(name);
     // Clone the tensor in the DataSet
-    TensorBase* tensor = this->_get_tensorbase_obj(name);
-    this->_tensor_memory.add_tensor(tensor);
+    TensorBase* tensor = _get_tensorbase_obj(name);
+    if (tensor == NULL) {
```
mellis13, line 84: Is there a code path to hit this?

src/cpp/commandlist.cpp
billschereriii: I disagree strongly on this one. Second rule of defensive programming is to assume that any other routine you call has the potential to give back a bad result -- whether through a bad input, a programming error, or an environmental issue of some sort. All return values from other functions must be sanity checked.

src/cpp/dataset.cpp
billschereriii: Again, never know when an error or other surprise can come up as code changes over time, so best practice is to code defensively

include/enums/c_tensor_type.h
billschereriii: fixed -- a tab got inserted instead of spaces

include/enums/cpp_metadata_type.h
billschereriii: done

include/enums/cpp_tensor_type.h
billschereriii: I'll change it to invalid when I consolidate the enumerations during the error handling work. Cases added to the switch statements

include/tensor.tcc
billschereriii: fixed

include/tensor.tcc
billschereriii: fixed. (These all came up from removing "this->", in case you're curious.)

include/enums/cpp_tensor_type.h
billschereriii: done

src/cpp/dbnode.cpp
```diff
@@ -30,29 +30,27 @@
...
+   // NOP
 }
 
-DBNode::DBNode(std::string ip, std::string name,
-               uint64_t port, uint64_t l_slot,
-               uint64_t u_slot, std::string prefix)
+// DBNode constructor with connection and hash slot information.
+DBNode::DBNode(std::string _ip, std::string _name, uint64_t _port,
+               uint64_t l_slot, uint64_t u_slot, std::string _prefix)
+    : ip(_ip), port(_port), name(_name), lower_hash_slot(l_slot),
```
mellis13, line 43: I don't think we should start passing in variables with ``_`` prefix because we have removed ``this``.  I think that's bad example to set for future developers.  Can we think of better solution?

src/cpp/gettensorcommand.cpp
```diff
@@ -31,8 +31,8 @@
 
 using namespace SmartRedis;
 
-const int num_elements = 2;
-const int num_fields = 6;
+const size_t num_elements = 2;
```
mellis13, line 34: Good catch here.  I think that ``type_idx``, ``dims_idx``, and ``blob_idx`` are probably better as ``size_t`` as well since ``operator[]`` accepts ``size_t``.

src/cpp/dbnode.cpp
billschereriii: I'm not sure where the confusion is -- this is standard practice for constructors to differentiate between the variable supplied to the constructor and the member variable. C++ scoping rules say that locally scoped variables have first priority for a name. Second priority (which is the case here) is the parameters to a function. In the case of initializers, there can't be any locally scoped variables, so these are the top priority. Let's discuss offline.

src/cpp/gettensorcommand.cpp
billschereriii: Yep. This was generating a warning about signed/unsigned and/or size mismatches once we turned on -Wall.

src/cpp/redisserver.cpp
```diff
@@ -26,74 +26,68 @@
...
  */
 
+#include <ctype.h>
 #include "redisserver.h"
 
 using namespace SmartRedis;
 
+// We should never seed srand more than once. There are more elegant ways
+// to prevent it, but this will suffice
+static bool ___srand_seeded = false;
```
mellis13, line 36: I think a better solution might be to make something like a ``std::default_random_engine`` as a member variable and then init that seed upon construction of redisserver.

src/cpp/tensorbase.cpp
```diff
@@ -42,130 +43,151 @@ TensorBase::TensorBase(const std::string& name,
...
+    tb._data = NULL;
 }
 
+// TensorBase destructor
 TensorBase::~TensorBase()
 {
-    if(this->_data)
-        free(this->_data);
+    if (_data != NULL) {
+        delete(reinterpret_cast<unsigned char *>(_data));
```
mellis13, line 86: I'm curious why the cast to ``unsigned char *``.  Doesn't ``delete`` take ``void*``?

billschereriii: I think I was getting a warning for something. I'll pull the cast and recompile to confirm

src/cpp/tensorbase.cpp
```diff
@@ -42,130 +43,151 @@ TensorBase::TensorBase(const std::string& name,
...
+// Retrieve the total number of values in the tensor.
 size_t TensorBase::num_values()
 {
-    /* Return the total number of values in the tensor
-    */
-    size_t n_values = this->_dims[0];
-    for(size_t i=1; i<this->_dims.size(); i++) {
-        n_values *= this->_dims[i];
-    }
+    size_t n_values = 1;
```
mellis13, line 166: On one hand I can see how this is better, but I think the previous is slightly better because an error would be thrown if ``_dims.size()`` is 0 (bad state), whereas the second would just return ``1`` which is not true and misleading.

src/cpp/tensorpack.cpp
```diff
@@ -30,152 +30,153 @@
...
 }
 
+// TensorPack copy assignment operator
 TensorPack& TensorPack::operator=(const TensorPack& tp)
 {
-    if(this!=&tp) {
-        this->_all_tensors.clear();
-        this->_tensorbase_inventory.clear();
-        this->_copy_tensor_inventory(tp);
+    if (this != &tp) {
```
mellis13, line 43: All others were ``if (this == &tp) return *this``.

src/cpp/tensorpack.cpp
```diff
@@ -30,152 +30,153 @@
...
-            ptr = new Tensor<uint16_t>(name, data, dims,
-                                       type, mem_layout);
+        case TensorType::uint16:
+            ptr = new Tensor<uint16_t>(name, data, dims, type, mem_layout);
             break;
-        case TensorType::uint8 :
-             ptr = new Tensor<uint8_t>(name, data, dims,
-                                       type, mem_layout);
+        case TensorType::uint8:
+            ptr = new Tensor<uint8_t>(name, data, dims, type, mem_layout);
```
mellis13, line 97: Alignment

src/cpp/tensorpack.cpp
```diff
@@ -30,152 +30,153 @@
...
+    return _tensorbase_inventory.at(name);
 }
 
+// Retrieve a pointer to the tensor data memory space
 void* TensorPack::get_tensor_data(const std::string& name)
 {
-    TensorBase* ptr = this->_tensorbase_inventory.at(name);
-    return ptr->data();
+    TensorBase* ptr = _tensorbase_inventory.at(name);
+    return (ptr == NULL ? ptr : ptr->data());
```
mellis13, line 132: I don't think this is the correct behavior.  Returning ``NULL`` implies that the data is ``NULL`` (since the function name is get_tensor_data()).  If we are worried that the value in the map is ``NULL``, then I think we should throw an error instead of return ``NULL``.

src/cpp/tensorpack.cpp
```diff
@@ -30,152 +30,153 @@
...
-        this->_tensorbase_inventory[ptr->name()] = ptr;
-        it++;
+    // Check for self-copy
+    if (this == &tp)
+        return;
+
+    typename TensorPack::const_tensorbase_iterator it = tp.tensor_cbegin();
+    for ( ; it != tp.tensor_cend(); it++) {
+        TensorBase* ptr = (*it)->clone();
+        if (ptr == NULL)
```
mellis13, line 175: We should check if clone() can return ``NULL``, and if it can, I think we should throw an error at the end of clone``clone()``.  Also, if ``(*it)`` is a valid tensor and we skip it because clone is null, the copy operation failed and we should throw an error, not skip it and have the tensorpacks be difference.

tests/cpp/unit-tests/test_commandlist.cpp
```diff
@@ -284,7 +284,7 @@ SCENARIO("Testing CommandList object on heap", "[CommandList]")
                 // Ensure that the state of the original
                 // CommandList object is preserved
                 while((it != it_end) && (cmd_count < fields.size())) {
-                    ((*it)->first_field() == fields[cmd_count]);
+                    CHECK((*it)->first_field() == fields[cmd_count]);
```
mellis13, line 287: Good catch!

src/cpp/client.cpp
```diff
@@ -30,386 +30,408 @@
...
-        type = GetTensorCommand::get_data_type(reply);
+        std::vector<size_t> reply_dims = GetTensorCommand::get_dims(reply);
+        std::string_view blob = GetTensorCommand::get_data_blob(reply);
+        TensorType type = GetTensorCommand::get_data_type(reply);
         dataset._add_to_tensorpack(tensor_names[i],
                                    (void*)blob.data(), reply_dims,
                                    type, MemoryLayout::contiguous);
     }
+
+    // ***WS*** FINDME: Passing objects by value causes unnecessary copies
```
mellis13, line 100: Was this resolved in this PR?  If not we can make a github issue for it.

src/cpp/client.cpp
```diff
@@ -30,386 +30,408 @@
...
+// Clone the dataset to a new name
 void Client::copy_dataset(const std::string& src_name,
                           const std::string& dest_name)
 {
-    CommandReply reply = this->_get_dataset_metadata(src_name);
-    DataSet dataset = DataSet(src_name);
-    this->_unpack_dataset_metadata(dataset, reply);
+    // Extract metadata
+    CommandReply reply = _get_dataset_metadata(src_name);
+    if (reply.n_elements() == 0) {
```
mellis13, line 118: Maybe this check would be better in ``_unpack_dataset_metadata`` so it gets used every time?

src/cpp/client.cpp
```diff
@@ -30,386 +30,408 @@
...
+            tensor = new Tensor<uint16_t>(get_key, (void*)blob.data(),
+                                          reply_dims, reply_type,
+                                          MemoryLayout::contiguous);
+            break;
+        case TensorType::uint8:
+            tensor = new Tensor<uint8_t>(get_key, (void*)blob.data(),
+                                         reply_dims, reply_type,
+                                         MemoryLayout::contiguous);
+            break;
         default : // FINDME: Need to handle this case better
```
mellis13, line 393: Just commenting here for bookkeeping.  If we need to add the previous ``FINDME`` as an issue, this one seems to match with it and could maybe be combined.

PR #147, opened 2021-09-13
Fix  in the Python client.get_dataset() return type.

mellis13:
Currently the python client returns the type PyDataset
(the pybind type) instead of the Dataset type.  
To fix this, the Dataset type no longer inherits from PyDataset, 
but Dataset has a PyDataset member variable.  
A test was written that sets a dataset, retrieves the dataset,
adds a tensor and metadata, sets the augmented dataset,
and then retrieves the augmented dataset.  While writing this test
for this return type fix, it was discovered that HDEL was not
being run before setting the metadata field for a DataSet.
This means that the HMSET would only be appending to the key
instead of clearing the key and then writing it.  The HDEL command
has been add to the put_dataset() command.

### DIFF
--- a/src/cpp/client.cpp
+++ b/src/cpp/client.cpp
@@ -850,7 +850,10 @@ void Client::_append_dataset_metadata_commands(CommandList& cmd_list,
                                  "a DataSet into the database that "\
                                  "does not contain any fields or "\
                                  "tensors.");
-    // SingleKeyCommand* cmd = new SingleKeyCommand();
+    SingleKeyCommand* del_cmd = cmd_list.add_command<SingleKeyCommand>();
+    del_cmd->add_field("DEL");
+    del_cmd->add_field(meta_key, true);
+
     SingleKeyCommand* cmd = cmd_list.add_command<SingleKeyCommand>();
     cmd->add_field("HMSET");
     cmd->add_field (meta_key, true);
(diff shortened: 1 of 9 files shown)


src/cpp/client.cpp
```diff
@@ -850,7 +850,10 @@ void Client::_append_dataset_metadata_commands(CommandList& cmd_list,
                                  "a DataSet into the database that "\
                                  "does not contain any fields or "\
                                  "tensors.");
-    // SingleKeyCommand* cmd = new SingleKeyCommand();
+    SingleKeyCommand* del_cmd = cmd_list.add_command<SingleKeyCommand>();
+    del_cmd->add_field("DEL");
+    del_cmd->add_field(meta_key, true);
```
billschereriii, line 855: We're creating a command and then doing nothing with it here -- doesn't seem to be intended

billschereriii: Was this test code that didn't get deleted?

mellis13: ``del_cmd`` is being added to the ``cmd_list`` to be executed

mellis13: I'm not following that "nothing is done with it".  Can you elaborate?

billschereriii: My bad, the semantics tripped me up again.

Spartee (approved):
Compiled and tested locally. Works as expected. Coverage was at 79%. Might be good to add in some error handling tests for the ``get_data`` and ``set_data`` functions.

PR #149, opened 2021-09-21
Build Fortran client in 'make lib'

ashao:
Users will now compile the Fortran client as part of the static
and shared SmartRedis libraries. This also creates a new directory
'mod' in the install directory that users are expected to add to
their application's compilation include flags.

NOTE: This implicitly assumes that the user's compilation environment has a Fortran 2018 compliant compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,9 @@ cmake_minimum_required(VERSION 3.10)
 
 option(BUILD_PYTHON "Build the python module" ON)
 
+enable_language(Fortran)
+set(CMAKE_Fortran_MODULE_DIRECTORY ${CMAKE_BINARY_DIR})
+
 set(CMAKE_BUILD_TYPE RELEASE)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_EXE_LINKER_FLAGS "-lpthread")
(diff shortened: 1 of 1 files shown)


ashao:
Thanks @billschereriii. I just tried it fresh again on cheyenne and it worked fine. I've got an Ubuntu 20.04 machine that I'll try next. I'll ping you when I resolve the problem.

billschereriii:
Sounds good. I've seen elsewhere that the test system is actually pickier than most others, so I fully believe that the change is working fine on your main system. It's a good change and well worth getting in, so please let me know if I can do anything to help

ashao:
Yah, no worries; picky testing systems are good. Only bit of info that might help me is if you happen to know what version of CMake the pipeline is using. I suspect this might be something picky about whether you include a trailing `/` for a directory or not.

ashao:
I looked at this a little more over the weekend and at least I got it to fail consistently. Apparently this is a known issue when trying to build multiple targets in parallel with Fortran modules. I'll have a time to dig into this today to fix this up.

billschereriii:
Any luck resolving the build issue here?

Spartee:
@ashao whats the progress here? We should close if this is a non-issue or re-investigate if it is.

ashao:
I'm going to close this for now until I get some time to look at it again.

PR #150, opened 2021-09-28
Add Dockerfile

gcoelho:
Dockerfile for SmartRedis containing

- SmartRedis static library

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,0 +1,51 @@
+#
+# BSD 2-Clause License
+#
+# Copyright (c) 2021, Hewlett Packard Enterprise
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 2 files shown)


images/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Get source code
+WORKDIR /usr/local/src/
+RUN git config --global advice.detachedHead false
+RUN git clone https://github.com/CrayLabs/SmartRedis.git
+
+# Compile and install
+WORKDIR /usr/local/src/SmartRedis
+RUN make lib
+
```
Spartee, line 48: So ``make lib`` builds the .so for the python library. I'm not sure exactly what the standard practice is for python package images, but I would think we have a couple options

1. manually place the ``smartredis/src/python/module/smartredis` dir in the ``PYTHONPATH``
2. run `pip install .`` (non-editable install)
3. run ``pip install -e .`` (editable install)

I tend to think option 2 would be the best, but option 1 would work across whatever python environments the user ends up creating in a derivative build (i.e. ``conda``)

images/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+#
+
+## Builder image
+FROM ubuntu:21.04 as builder
+
+# Install tooling
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -q -y --no-install-recommends \
+    ca-certificates git \
+    make cmake gcc g++ libc6-dev python3 libpython3-dev pip
```
Spartee, line 36: we should add a `` && rm -rf /var/lib/apt/lists/*`` line here to reduce the image size.

Spartee (changes requested):
Thanks for the contribution @gcoelho! I had some minor comments. 

I'm not exactly sure what the best practices are for installing a python library into a docker image that is meant to be included in a build but I listed some out. 

After the Python client is installed I will approve this PR.

Thanks again!

mellis13:
Closing this PR because work will continue on this via #180

PR #165, opened 2021-11-13
Add CI for all SmartRedis tests

Spartee:
# Description

Pull request for enabling the CI (github actions) on github instead of jenkins. 

### testing Implementation
 - macOS not tested explicitly as docker commands not supported in macOS github actions
 - uses RedisAI service container which can be switched based on versions
 - only runs non-cluster tests
 - runs all unit tests and tests for all languages in one job (probably not optimal)

Changes to tests
 - run command (i.e. mpirun) taken out of tests so we don't have that testing dependency

# Helpful resources
 - [Python Setup action](https://github.com/actions/setup-python)
 - [Redis Service containers](https://docs.github.com/en/actions/using-containerized-services/creating-redis-service-containers)
 - [Apt timeouts](https://github.com/actions/virtual-environments/issues/675)
 - [CodeCov](https://about.codecov.io/)
 - [Ci for Pandas](https://github.com/pandas-dev/pandas/blob/master/.github/workflows/ci.yml)

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -27,9 +27,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, macos-10.15]
-        gcc_v: [8] # Version of GFortran we want to use.
-        rai: [1.2.4] # add 1.2.5 in the future
+        os: [ubuntu-20.04] # cannot test on macOS as docker isn't supported on Mac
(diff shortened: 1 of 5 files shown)


Spartee:
@mellis13 please see testing changes and make sure that works for you. Taking out the run command will mean we need to change how we suggest testing but overall I think it's a good idea.

@EricGustin this should be the majority of what is needed to complete #165. There are two TODO items left.

Issue #166, opened 2021-11-16
fortran_c_interop bug in transfer statement 


jedwards4b:
# Description
The transfer statement at line 74 of fortran_c_interop.F90 produces an error if length(character_array_f(i)) == 0 when using the intel 19.1.1 compiler.   The error is: When the SOURCE argument to the intrinsic function TRANSFER has size greater than zero, the MOLD argument must not be an array with elements of size zero.

# How to reproduce
steps to reproduce the bug  

# Expected behavior

Here is a proposed solution:
```
@@ -71,11 +71,13 @@ subroutine convert_char_array_to_c(character_array_f, character_array_c, string_
 

   ! Copy the character into a c_char and create pointers to each of the strings
   do i=1,n_strings
-    character_array_c(i) = transfer(character_array_f(i),character_array_c(i))
-    string_ptrs(i) = c_loc(character_array_c(i))
+     if(string_lengths(i) > 0) then
+        character_array_c(i) = transfer(character_array_f(i),character_array_c(i))
+        string_ptrs(i) = c_loc(character_array_c(i))
+     endif


```
# System
 - OS: Linux
 - OS version:
 - Commit/version of library: v0.2.0

ashao:
I agree with the proposed solution.

billschereriii:
I agree with the solution and with the tweak

billschereriii:
Addressed in PR #170

billschereriii:
@jedwards4b I couldn't list you as a formal reviewer, but if you could look at the PR, I'd appreciate your feedback.

PR #170, opened 2021-11-17
Fix zero-length array transfer in convert_char_array_to_c

billschereriii:
Address #166 by checking length of array. Assign null if array is length zero.


jedwards4b (approved):

jedwards4b:
@billschereriii thanks!

ashao (approved):
Looks good to me. Thanks!

PR #177, opened 2021-12-03
Build matrix for CI

EricGustin:
This PR implements the build matrix for the public CI on GitHub Actions for SmartRedis. The matrix has 24 combinations and looks like:

**OS:** ubuntu-20.04
**Compiler:** intel, GCC/GFortran 8, GCC/GFortran 9, GCC/GFortran 10
**RedisAI versions:** 1.2.4, 1.2.5
**Python versions:** 3.7.x, 3.8.x, 3.9.x

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -22,18 +22,18 @@ env:
 jobs:
 
   run_tests:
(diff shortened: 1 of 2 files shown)


EricGustin:
See PSD issue 110 for more info.

Spartee (approved):
Approving as I want this to go in. See comment on Ubuntu 18 error. I think it's fine we don't cover two ubuntu operating systems honestly, they will be very similar. But we should add it to the nightly we end up doing.

PR #180, opened 2021-12-09
Dockerfile

mellis13:
This PR builds on the work from #150 to create a Dockerfile for SmartRedis.  A couple of changes have been made from the original PR:

- The ``include`` directories are copied to ``/usr/local/include`` along with the ``lib`` directory to ``/usr/local/lib``
- In the original PR, a sequential build was used, with the last image being a clean derivative of ``ubuntu:21.04``.  In that derivative image, only the smartredis lib was copied over.  This leaves behind any pip install of the smartredis client as well as header files.  For now, the sequential build has been removed so that the installed smartredis python package and compiled client library are both accessible.

The final stage of the Dockerfile may need to be refiend to minimize image size and provide maximum extensibility for derivative images.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -29,6 +29,10 @@
 ## Builder image
 FROM ubuntu:21.04 as builder
 
+# Define default arguments for building
+ARG REMOTE=https://github.com/CrayLabs/SmartRedis.git
+ARG BRANCH=develop
+
 # Install tooling
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -q -y --no-install-recommends \
@@ -41,7 +45,7 @@ RUN pip3 install cmake
 # Get source code
 WORKDIR /usr/local/src/
 RUN git config --global advice.detachedHead false && \
-    git clone https://github.com/CrayLabs/SmartRedis.git
+    git clone $REMOTE --branch $BRANCH
 
 # Compile and install
 WORKDIR /usr/local/src/SmartRedis
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -33,7 +33,8 @@ FROM ubuntu:21.04 as builder
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -q -y --no-install-recommends \
     ca-certificates git \
-    make cmake gcc g++ libc6-dev python3 libpython3-dev pip
+    make cmake gcc g++ libc6-dev python3 libpython3-dev pip && \
+    rm -rf /var/lib/apt/lists/*
 RUN ln -s /usr/bin/python3 /usr/bin/python
 RUN pip3 install cmake
 
@@ -44,8 +45,11 @@ RUN git clone https://github.com/CrayLabs/SmartRedis.git
 
 # Compile and install
 WORKDIR /usr/local/src/SmartRedis
-RUN make lib
+RUN make lib && pip install .
 
-## Runtime image
-FROM ubuntu:21.04
-COPY --from=builder /usr/local/src/SmartRedis/install/lib/libsmartredis.so /usr/local/lib/
+# Copy install files and directories to /usr/local/lib and
+# usr/local/include and delete unnecessary build files
+RUN cp -r /usr/local/src/SmartRedis/install/lib/* /usr/local/lib/ && \
+    mkdir /usr/local/include/smartredis && \
+    cp -r /usr/local/src/SmartRedis/install/include/* /usr/local/include/smartredis/ && \
+    rm -r /usr/local/src/SmartRedis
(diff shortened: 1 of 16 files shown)


mellis13:
For added context on container usage, the SmartRedis Dockerfile can be used to build a derivative containers such as:

```
FROM smartredis

COPY ./CMakeLists.txt /usr/local/src/DockerTest/
COPY ./test_docker.cpp /usr/local/src/DockerTest/
WORKDIR /usr/local/src/DockerTest
RUN mkdir build && cd build && cmake .. && make && cd ../
```

where ``test_docker.cpp`` is any C++ application using the SmartRedis client and CMakeLists.txt is the build file for the application:

```c++
#include "client.h"
#include <vector>
#include <string>

int main(int argc, char* argv[]) {

  SmartRedis::Client c(false);
  return 0;
}
```

```
project(DockerTester)

cmake_minimum_required(VERSION 3.10)

set(CMAKE_VERBOSE_MAKEFILE ON)
set(CMAKE_BUILD_TYPE DEBUG)
set(CMAKE_CXX_STANDARD 17)

find_library(SR_LIB smartredis)

include_directories(SYSTEM
    /usr/local/include/smartredis
)

# Build executables

add_executable(docker_test
	test_docker.cpp
)
target_link_libraries(docker_test ${SR_LIB} pthread)
```

images/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+FROM ubuntu:21.04 as builder
+
+# Install tooling
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -q -y --no-install-recommends \
+    ca-certificates git \
+    make cmake gcc g++ libc6-dev python3 libpython3-dev pip && \
+    rm -rf /var/lib/apt/lists/*
+RUN ln -s /usr/bin/python3 /usr/bin/python
+RUN pip3 install cmake
```
Spartee, line 39: given the pip based cmake issues, wondering if we should just apt install it? We would have to take out the cmake requirement for SmartRedis but now might be a good time to do that. pinging @al-rigazzi  on this since hes seen this problems most recently.

images/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+# Get source code
+WORKDIR /usr/local/src/
+RUN git config --global advice.detachedHead false && \
+    git clone https://github.com/CrayLabs/SmartRedis.git
```
Spartee, line 44: ``--branch``?

tests/docker/c/test_docker.c
```diff
@@ -0,0 +1,63 @@
...
+    char* key = "c_docker_tensor";
+    size_t key_length = strlen(key);
+
+    double tensor[3] = {1.0, 2.0, 3.0};
+    SRTensorType type = SRTensorTypeDouble;
+    SRMemoryLayout layout = SRMemLayoutContiguous;
+    size_t dims[1] = {3};
+    size_t n_dims = 1;
+
+    return_code = put_tensor(client, key, key_length,
```
Spartee, line 55: should we also get the tensor and compare the arrays?

Spartee (approved):
Looks really great. Love that we got to this. A couple things

1. We should instructions to the docs that specify how to build your application into a container with SmartRedis. You basically already have this.
2. Tests should probably get the tensor and compare.
3. see cmake question.

Leaving up to you if you want to merge and then address these. Works on my machine.

tests/docker/c/test_docker.c
mellis13: Good idea -- I added equivalence check to each test.

images/Dockerfile
mellis13: I think we should take out the pip install cmake requirement, but maybe that should be tracked as a separate issue and PR for better visibility (and easier to unwind if we end up changing our minds).

images/Dockerfile
mellis13: Yeah, this brings up a good point.  By default this will clone from ``develop`` which is not what we want for the github actions.  I attempted to get a ``--branch`` option to work with github actions via docker ``--build-args``, but this ended up being problematic with the way that github actions does the branch and sha names differently for ``push`` actions and ``pull request`` actions.  As a result, I've left a comment in the Dockerfile for the user (with code) on how to clone a specific repository and branch for use in the Dockerfile, but otherwise the default behavior will be to build from the current state of the repo in which ``./images/build.sh`` is invoked.  This guarantees that our github actions are executed correctly for both push and pull request actions.  After doing some research, I have seen this as accepted behavior (and often preferred) so that developers can make and test changes to the project without committing and pushing changes.  I'll make sure to document this in the doc updates associated with this PR.

images/Dockerfile
Spartee: @al-rigazzi is doing this.

PR #183, opened 2022-01-05
Move -Werror and -Wall to test build

mellis13:
This PR moves the ``-Werror`` and ``-Wall`` compile flags only to the test build.  The optional flags can be added via the ``WERROR`` CMake variable.  This variable is set initially in the library ``Makefile`` and passed through ``build_lib.sh``.   It was verified that the flags work correctly by injecting an unused variable into the SmartRedis code and observing that the library builds but the test suite does not.


billschereriii:
This looks good so far. Do we need to add WERROR to CI to complete this PR?

mellis13:
Good question -- I think since ``run_tests.yml`` executes ``make test-verbose``, the test suite built and run with the CI should have ``WERROR`` defined.

billschereriii:
Great PR, thanks for taking care of this!

billschereriii (approved):

PR #189, opened 2022-01-13
Fix CMake issues with CUDNN

al-rigazzi:
This PR aims at removing `pip`-installed version of CMake, as it was causing problems for users when building SmartRedis and SmartSim from source.

In detail, wherever we were using the `cmake` python module to find the `cmake` binary, now we just run `which cmake`. This is compatible with whatever CMake is available system-wide. In our Github actions, we `brew`/`apt-get` install CMake instead of `pip`-installing it. The same should happen for `PyPI`.

Users who use `conda` envs, will have the possibility to run `conda install cmake` and build both SmartRedis and SmartSim from source.


mellis13 (commented):
This will be a great improvement and solve a lot of confusion on the build side.  Just a couple of questions/comments:

- Do you think we should remove ``cmake>=3.10`` from ``requirements.txt``?
- I think that ``RUN pip3 install cmake && rm -rf ~/.cache/pip`` in images/Dockerfile can now be removed (we apt-get install it above that line)

al-rigazzi:
@mellis13 
- yes, I removed `cmake==3.10` from the `requirements.txt` and `requirements-dev.txt` files. The minimum version is already correctly required in `CMakeLists.txt`
- right, removed that line too, good catch, thanks!

mellis13 (approved):
Thanks for the PR!  This will eliminate a lot of build confusion.  Looks good to merge.

PR #195, opened 2022-01-24
Dataset acknowledgement key refactor

mellis13:
Previously the ``Client`` class stored an acknowledgement key at the key ``dataset_name``.  However, tensor and metadata keys for the dataset were stored at ``{dataset_name}.tensor_name`` and ``{dataset_name}.meta``.  This is OK for non-ensemble use cases, but in an ensemble, the keys take the form:

Ack key: ``ensemble_name.dataset_name``
Tensor keys: ``ensemble_name.{dataset_name}.tensor_name``
Metadata key: ``ensemble_name.{dataset_name}.meta``.

This means that for ensembles the ack key will hash to a different slot than the tensors and metadata.  This prevents pipelining dataset ``HSET`` and ``AI.TENSORSET`` operations into a single roundtrip communication.    

This PR changes the acknowledgement key to be a hash field in the ``.meta`` key.  In this way, the ack key is easily included and pipelined in the dataset ``HSET`` and ``AI.TENSORSET`` operations.

### DIFF
--- a/src/cpp/client.cpp
+++ b/src/cpp/client.cpp
@@ -63,7 +63,7 @@ Client::~Client()
     _redis_server = NULL;
 }
 
-// Establish a dataset
+// Put a DataSet object into the database
 void Client::put_dataset(DataSet& dataset)
 {
     CommandList cmds;
@@ -73,7 +73,7 @@ void Client::put_dataset(DataSet& dataset)
     _run(cmds);
 }
 
-// Retrieve the current dataset
+// Retrieve a DataSet object from the database
 DataSet Client::get_dataset(const std::string& name)
 {
     // Get the metadata message and construct DataSet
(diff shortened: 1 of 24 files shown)


src/cpp/client.cpp
```diff
@@ -624,8 +616,8 @@ bool Client::tensor_exists(const std::string& name)
 bool Client::dataset_exists(const std::string& name)
 {
     // Same implementation as for tensors; the next line is NOT a type
```
al-rigazzi, line 618: I'm a bit puzzled by this comment

al-rigazzi (approved):
Good work! I like this approach much better than the previous one. LGTM (I only added one comment to a comment, which was not actually part of this PR).

src/cpp/client.cpp
mellis13: yes, this old comment should be removed.  Also fixed the typo above the ``dataset_exists()`` function name.

PR #197, opened 2022-01-26
Add AI.INFO command to Python and C++ clients

mellis13:
The AI.INFO command has been added to the Python and C++ clients.  The commands allow users to retrieve statistics on scripts and models (metadata and execution information).  The API functions follow the same format as other "information" commands:  users are required to give the address of the database nodes where the command is to be executed.  Like the other "information" commands, the Python client supports multiple addresses.  The results are returned in Python as a list of dictionaries.

### DIFF
--- a/src/cpp/client.cpp
+++ b/src/cpp/client.cpp
@@ -784,6 +784,48 @@ parsed_reply_map Client::get_db_cluster_info(std::string address)
                                                      reply.str_len()));
 }
 
+// Returns the AI.INFO command reply addressed to any database shard
+parsed_reply_map Client::get_ai_info()
+{
+    // Run the AI.INFO command
+    AddressAnyCommand cmd;
+    cmd.add_field("AI.INFO", true);
+    CommandReply reply = _run(cmd);
+
+    if (reply.has_error())
+        throw SRRuntimeException("AI.INFO command failed on server");
+
+    if (reply.n_elements() % 2 != 0)
+        throw SRInternalException("The AI.INFO reply structure has an "\
+                                  "unexpected format");
+
+    // Parse reply
+    parsed_reply_map reply_map;
+    std::string key;
+    std::string value;
+    for(size_t i = 0; i < reply.n_elements(); i += 2){
+
+        if(reply[i].redis_reply_type() == "REDIS_REPLY_STRING")
+            key = reply[i].str();
+        else if (reply[i].redis_reply_type() == "REDIS_REPLY_STATUS")
+            key = reply[i].status_str();
+        else
+            throw SRInternalException("The AI.INFO reply does not have "\
+                                      "the correct type.");
+
+        if(reply[i+1].redis_reply_type() == "REDIS_REPLY_STRING")
+            value = reply[i+1].str();
+        else if (reply[i+1].redis_reply_type() == "REDIS_REPLY_STATUS")
+            value = reply[i+1].status_str();
+        else
+            throw SRInternalException("The AI.INFO reply does not have "\
+                                      "the correct type.");
+
+        reply_map[key] = value;
+    }
+    return reply_map;
+}
+
 // Delete all the keys of the given database
 void Client::flush_db(std::string address)
 {
--- a/src/cpp/client.cpp
+++ b/src/cpp/client.cpp
@@ -784,13 +784,14 @@ parsed_reply_map Client::get_db_cluster_info(std::string address)
                                                      reply.str_len()));
 }
 
-// Returns the AI.INFO command reply addressed to any database shard
-parsed_reply_map Client::get_ai_info()
+// Returns the AI.INFO command reply
+parsed_reply_map Client::get_ai_info(const std::string& address,
+                                     const std::string& key,
+                                     const bool reset_stat)
 {
-    // Run the AI.INFO command
-    AddressAnyCommand cmd;
-    cmd.add_field("AI.INFO", true);
-    CommandReply reply = _run(cmd);
+    // Run the command
+    CommandReply reply =
+        _redis_server->get_model_script_ai_info(address, key, reset_stat);
 
     if (reply.has_error())
         throw SRRuntimeException("AI.INFO command failed on server");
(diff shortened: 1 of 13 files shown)


src/cpp/client.cpp
```diff
@@ -784,6 +784,43 @@ parsed_reply_map Client::get_db_cluster_info(std::string address)
...
+    for (size_t i = 0; i < reply.n_elements(); i += 2) {
+        std::string map_key = reply[i].str();
+        std::string value;
+        if(reply[i+1].redis_reply_type() == "REDIS_REPLY_STRING")
+            value = reply[i+1].str();
+        else if (reply[i+1].redis_reply_type() == "REDIS_REPLY_INTEGER")
+            value = std::to_string(reply[i+1].integer());
+        else if (reply[i+1].redis_reply_type() == "REDIS_REPLY_DOUBLE")
+            value = std::to_string(reply[i+1].dbl());
+        else
```
billschereriii, line 816: space out "i+1" to "i + 1" (6x)

src/cpp/rediscluster.cpp
```diff
@@ -569,6 +569,41 @@ CommandReply RedisCluster::get_script(const std::string& key)
...
+    uint64_t port = cmd.parse_port(address);
+
+    // Determine the prefix we need for the model or script
+    if (!is_addressable(host, port)) {
+        throw SRInternalException("The provided host and port does "\
+                                  "not match a cluster shard address.");
+    }
+
+    std::string host_port = host + ":" + std::to_string(port);
+    std::string db_prefix = _address_node_map.at(host_port)->prefix;
```
billschereriii, line 590: Why not just use the address string supplied by the caller here?

src/cpp/rediscluster.cpp
```diff
@@ -569,6 +569,41 @@ CommandReply RedisCluster::get_script(const std::string& key)
...
+
+    std::string prefixed_key = "{" + db_prefix + "}." + key;
+
```
billschereriii, line 593: Isn't there a utility routine for this?

src/python/module/smartredis/client.py
```diff
@@ -713,6 +713,28 @@ def get_db_cluster_info(self, addresses):
...
+                  list corresponding to an address reply
+        :rtype: list[dict]
+        :raises RedisReplyError: if there is an error
+                in command execution or parsing the command reply.
+        """
+        try:
+            return super().get_ai_info(address, key, reset_stat)
+        except RuntimeError as e:
+            raise RedisReplyError(str(e), "get_ai_info()")
+
```
billschereriii, line 737: This will need to be rewritten to match the new approach from the python error handling PR. This will mean:
-  Adding the `exception_handler` decorator`
- Performing type validation on the inputs to throw TypeError if anything is passed of incorrect type
- Adding type validation checks to the python unit tests

src/python/src/pyclient.cpp
```diff
@@ -601,6 +601,32 @@ std::vector<py::dict> PyClient::get_db_cluster_info(std::vector<std::string> add
...
+            ai_info.push_back(result_dict);
+        }
+        catch(const std::exception& e) {
+            throw SRRuntimeException(e.what());
+        }
+        catch(...) {
+            throw SRRuntimeException("A non-standard exception was encountered "\
+                                     "during client get_ai_info() execution.");
+        }
+    }
```
billschereriii, line 626: Put the whole function in the try/catch block, not just the interior loop

tests/python/test_nonkeyed_cmd.py
```diff
@@ -61,6 +61,31 @@ def test_dbcluster_info_command(use_cluster):
...
+    # Check with valid address and model key
+    ai_info = client.get_ai_info(address, "ai_info_cnn")
+    assert len(ai_info) != 0
+
+    # Check that invalid address throws error
+    with pytest.raises(RedisReplyError):
+        client.get_ai_info(["no_host:6379"], "ai_info_cnn")
+
+    # Check that invalid model name throws error
+    with pytest.raises(RedisReplyError):
```
billschereriii, line 87: Should be checking here for RedisRuntimeError

tests/python/test_nonkeyed_cmd.py
```diff
@@ -61,6 +61,31 @@ def test_dbcluster_info_command(use_cluster):
...
+    client.set_model("ai_info_cnn", model, "TORCH", "CPU")
+
+    # Check with valid address and model key
+    ai_info = client.get_ai_info(address, "ai_info_cnn")
+    assert len(ai_info) != 0
+
+    # Check that invalid address throws error
+    with pytest.raises(RedisReplyError):
+        client.get_ai_info(["no_host:6379"], "ai_info_cnn")
+
```
billschereriii, line 85: Should be checking here for RedisRuntimeError

mellis13: In ``client.py`` I am seeing that all ``RuntimeError`` are being reraised as ``RedisReplyError``.  That seems to be the current standard.  I'm also not seeing a ``RedisRuntimeError`` in ``error.py`` -- all current python exceptions seem to  inherit from ``RuntimeError``

al-rigazzi (approved):
Thanks for this PR, it will make developers' life much easier!

LGTM

src/cpp/rediscluster.cpp
mellis13: No, there is not a utility for this ``rediscluster.cpp`` that I am aware of.

tests/python/test_nonkeyed_cmd.py
billschereriii: That changes with the Python Error Handling PR --- sorry, should have been more clear

src/cpp/rediscluster.cpp
mellis13: ``AddressAtCommand::parse_host`` and ``AddressAtCommand::parse_port`` do error checking and reraise as SmartRedis errors.  This is the same paradigm we use for other ``AddressAtCommand``.  This flow could probably be improved, but that would require refactoring other commands in ``client.py`` that follow this pattern, and this seems beyond this ticket scope.

src/python/module/smartredis/client.py
mellis13: Agreed.  This PR will probably get merged first, so it will be up to the error handling PR to update this.

tests/python/test_nonkeyed_cmd.py
mellis13: Ah, makes more sense.  Thanks!

PR #201, opened 2022-01-27
Prevent Make Clean From Clobbering Catch2

MattToast:
Super quick and dirty solution to keep Catch2 header file from being removed on `make clean`. Instead of downloading the single header file, we clone the repo and use the header file provided there.

This also prepares us for the next major release where catch plans to move away from the single header file model. We will simply need to bump the tag and build.

### DIFF
--- a/build-scripts/build-catch.sh
+++ b/build-scripts/build-catch.sh
@@ -6,10 +6,7 @@ if [[ -f ./catch/single_include/catch2/catch.hpp ]]; then
 else
     echo "Installing Catch"
     if [[ ! -d "./catch" ]]; then
-        git clone https://github.com/catchorg/Catch2.git catch
-        cd catch
-        git checkout tags/v2.13.6
-        cd ..
+        git clone https://github.com/catchorg/Catch2.git catch --branch v2.13.6 --depth=1
     else
         echo "Catch downloaded"
     fi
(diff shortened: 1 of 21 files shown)


build-scripts/build-catch.sh
```diff
@@ -1,12 +1,15 @@
...
 
 # Install Catch
-if [[ -f ./catch/catch.hpp ]]; then
+if [[ -f ./catch/single_include/catch2/catch.hpp ]]; then
     echo "Catch has already been download and installed"
 else
     echo "Installing Catch"
     if [[ ! -d "./catch" ]]; then
-        wget https://github.com/catchorg/Catch2/releases/download/v2.13.6/catch.hpp -P catch
+        git clone https://github.com/catchorg/Catch2.git catch
```
mellis13, line 9: It's a minor optimization, but it is probably more efficient to just pull down what we need with a shallow clone:

```bash
git clone https://github.com/catchorg/Catch2.git catch --branch v2.13.6 --depth=1
```

mellis13 (approved):
One small suggestion, but otherwise looks great.  Thanks!

build-scripts/build-catch.sh
MattToast: Updated in cfe329a

PR #203, opened 2022-02-01
Streamline run() and connect() commands

billschereriii:
Streamline `run` and `connect` commands for Redis and RedisCluster to make sure that retries happen when they should and appropriate messages are generated when operations fail.


billschereriii:
fixes #203

mellis13 (approved):
Good cleanup -- LGTM

PR #208, opened 2022-02-09
Update copyright years: 2021 -> 2021-2022

ben-albrecht:
For future reference (in case we don't automate this by 2023):

Files updated via the following command:

    rpl -v -- "2021" "2021-2022" $(rg "\(c\) 2021" -l)

and then updated `doc/conf.py` to 2022 since there seems to be an established convention of using the latest year in documentation copyrights

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 BSD 2-Clause License
 
-Copyright (c) 2021, Hewlett Packard Enterprise
+Copyright (c) 2021-2022, Hewlett Packard Enterprise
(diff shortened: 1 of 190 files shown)


ben-albrecht:
Mentioning https://github.com/CrayLabs/SmartSim/pull/141 so these PRs are easier to find from each other, if we reference them in 2023 update.

ben-albrecht:
> Not sure I understand the reason behind the different form for doc/conf.py. The rest of the changes look good.

Updated the PR msg to explain this.

PR #211, opened 2022-02-12
Release documentation update

mellis13:
Documentation has been updated for the 0.3.0 release.  Updates include:

- Changing version numbers from 0.2.0 to 0.3.0
- Updating API examples to reflect improvements in use of ``key`` and ``name``
- Adding runtime environment variable content
- Adding python method names to summary table
- Update contact information

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -22,7 +22,7 @@ copyright = '2022, Hewlett Packard Enterprise'
 author = 'HPE AI & Advanced Productivity'
 
 # The full version, including alpha/beta/rc tags
-release = '0.2.0'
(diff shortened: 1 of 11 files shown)


doc/conf.py
```diff
@@ -22,7 +22,7 @@
 author = 'HPE AI & Advanced Productivity'
```
al-rigazzi, line 22: Is this still valid?

al-rigazzi (approved):
Just one question about the authorship, even though it is not related to this PR. The rest LGTM, the distinction between key and name also makes the docs more readable.

PR #214, opened 2022-02-15
Use C++11 override keyword

billschereriii:
Update the exception hierarchy to use the C++11 `override` keyword


billschereriii:
This generated a warning on theta

mellis13 (approved):
Thanks for fixing the compiler warning.  Which C++ compiler raised the warning?  It might we worth considering adding to our test matrix in a future PR.

LGTM

PR #215, opened 2022-02-15
Fix name conflict in dataset.h

billschereriii:
The name of the dataset was stored in the variable "name", but this conflicted with the parameter "name" used in several of the dataset member functions. Renamed the parameter to "_name" to help show it as a member variable and eliminate the conflict.


mellis13 (approved):
Looks great -- thanks!

PR #217, opened 2022-02-16
Update CMake minimum version to 3.13

mellis13:
The SmartRedis build process uses the CMake function ``add_link_options()`` which was added in version 3.13 (https://cmake.org/cmake/help/git-master/command/add_link_options.html).  As a result, we should bump our minimum version from 3.10 to 3.13.  The current version of CMake is 3.23, so users are unlikely to encounter an error unless they are using a very old version of CMake, and newer versions of CMake are easily installed in a ``conda`` environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 project(SmartRedis)
 
-cmake_minimum_required(VERSION 3.10)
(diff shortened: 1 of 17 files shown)


billschereriii (approved):

billschereriii:
LGTM, thanks for taking care of this!

PR #218, opened 2022-02-16
Make sure that CPP test coverage includes all source files

billschereriii:
Generate an error if the .cpp SOURCES list is found to have a different number of files in it than the number of .cpp files in src/cpp. This will ensure that our test coverage stays accurate when new files are added to the library.

### DIFF
--- a/tests/cpp/unit-tests/CMakeLists.txt
+++ b/tests/cpp/unit-tests/CMakeLists.txt
@@ -93,8 +93,8 @@ file(GLOB UNIT_TEST_FILES *.cpp)
 list(LENGTH UNIT_TEST_FILES cpp_unit_test_file_count)
 
 if(NOT (unit_test_count EQUAL cpp_unit_test_file_count))
-	message(FATAL_ERROR "CPP test coverage sources count (${unit_test_count}) \
-does not match count of cpp source files: ${cpp_unit_test_file_count}")
+	message(FATAL_ERROR "CPP unit test count (${unit_test_count}) \
+does not match count of cpp unit test files: ${cpp_unit_test_file_count}")
 endif()
 
 add_executable(cpp_unit_tests ${SOURCES} ${UNIT_TESTS})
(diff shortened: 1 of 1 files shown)


tests/cpp/unit-tests/CMakeLists.txt
```diff
@@ -65,6 +65,21 @@ set(UNIT_TESTS
...
+# coverate completely. It is not recommended to use globbed file lists
+# in add_executable()
+list(LENGTH SOURCES source_count)
+
+file(GLOB cpp_files ../../../src/cpp/*.cpp)
+
+list(LENGTH cpp_files cpp_file_count)
+
+if(NOT (source_count EQUAL cpp_file_count))
+	message(FATAL_ERROR "CPP test coverage sources (${source_count}) does not match \
```
EricGustin, line 79: Ignore this if you want, but mentioning that the first number for the CPP test coverage sources is the count might make the message more clear e.g. "CPP test coverage source count ${source_count}..." or "The count of CPP test coverage sources ${source_count}..."

EricGustin (approved):
This PR looks great. It will definitely serve as a safeguard against developers forgetting to update cpp coverage report details. 

I left a single comment about possible changes to the error message.

Since we now have this for the source file count, I think it would be beneficial to add this same functionality to the test file count. If you think this addition is beyond the scope of this PR, then we could talk about making a new ticket for that case.

tests/cpp/unit-tests/CMakeLists.txt
```diff
@@ -65,6 +67,36 @@ set(UNIT_TESTS
...
+# not running all the tests! It is not recommended to use globbed file lists
+# in add_executable()
+list(LENGTH UNIT_TESTS unit_test_count)
+
+file(GLOB UNIT_TEST_FILES *.cpp)
+
+list(LENGTH UNIT_TEST_FILES cpp_unit_test_file_count)
+
+if(NOT (unit_test_count EQUAL cpp_unit_test_file_count))
+	message(FATAL_ERROR "CPP test coverage sources count (${unit_test_count}) \
```
EricGustin, line 96: We should probably have a different error message than the sources message. Make sure this message references test file count rather than cpp source file count. e.g. "Target unit test file count does not match actual unit test file count"

EricGustin (changes requested):

EricGustin (approved):

PR #220, opened 2022-02-19
Run examples in CI

billschereriii:
- Update examples to provide a zero exit code on successful run
- Add pytest script to run all examples
- Add make target for `make test-examples`
- Add run of `make test examples` to CI script

### DIFF
--- a/.github/workflows/run_one_off.yml
+++ b/.github/workflows/run_one_off.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - master
       - develop
+#  pull_request:
+#    branches:
+#      - master
+#      - develop
 
 env:
   HOMEBREW_NO_ANALYTICS: "ON" # Make Homebrew installation a little quicker
@@ -24,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04] # cannot test on macOS as docker isn't supported on Mac
-        compiler: [intel, 8, 9, 10, 11] # intel compiler, and versions of GNU compiler
+        compiler: [8, 9, 10, 11] # intel compiler, and versions of GNU compiler
         rai_v: [1.2.4, 1.2.5] # verisons of RedisAI
         py_v: ['3.7.x', '3.8.x', '3.9.x'] # versions of Python
     env:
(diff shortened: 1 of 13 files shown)


billschereriii:
This PR is blocked by [Add clustered Redis testing to SmartRedis CI](https://github.com/CrayLabs/Poseidon/issues/164).

Everything is working except for the fact that the examples are written for clustered Redis and CI uses non-clustered.

Spartee (changes requested):
These should be in a different testing file (yml file) from ``run_test.yml``. Given that SR already has a whopping 34 checks for each check-in, lets run these once ever merged PR.

Changes:
- Separate workflow file
- run once every PR to develop.

.github/workflows/run_one_off.yml
```diff
@@ -0,0 +1,141 @@
+name: run_one_off
```
ben-albrecht, line 1: I recommend a name that reflects the contents or frequency of the test suite (if we expect the contents to change often), e.g. `run_examples`,  `run_post-merge_tests` (follows `run_tests` naming, or `post-merge_tests`

ben-albrecht (commented):
CI and Makefile changes LGTM, barring one inline comment

examples/test_examples.py
```diff
@@ -0,0 +1,101 @@
...
+
+@pytest.mark.parametrize("test", get_test_names())
+def test_cpp_client(test, use_cluster):
+    cmd = []
+    cmd.append(test)
+    print(f"Running test: {osp.basename(test)}")
+    print(f"Test command {' '.join(cmd)}")
+    print(f"Using cluster: {use_cluster}")
+    execute_cmd(cmd)
+    time.sleep(1)
```
ben-albrecht, line 55: Why do we need this sleep?

ben-albrecht (approved):
Rest of the changes LGTM.

examples/test_examples.py
billschereriii: Honestly, I'm not sure. It was in the other test driver scripts so I kept it there

ben-albrecht: It would be nice to understand the rationale and document it as a comment here for our future selves, but that's a nit (not PR-blocking).

.github/workflows/run_one_off.yml
billschereriii: renamed to run_post_merge_tests.yml

examples/test_examples.py
billschereriii: @Spartee Do you happen to know?

Spartee: on shared systems sometimes it can take an extra couple seconds to spin up

Spartee (approved):
Thanks for those changes. this looks good.

PR #223, opened 2022-02-25
Cmd refactor

### LINKED ISSUE #222: Support builder semantics for command construction
# Description
Add support for `operator<<()` to the various Command classes to facilitate easy construction of
Redis commands.

For example, this will allow replacement of:
```cpp
// Put a Tensor on the server
CommandReply Redis::put_tensor(TensorBase& tensor)
{
    // Build the command
    SingleKeyCommand cmd;
    cmd.add_field("AI.TENSORSET");
    cmd.add_field(tensor.name());
    cmd.add_field(tensor.type_str());
    cmd.add_fields(tensor.dims());
    cmd.add_field("BLOB");
    cmd.add_field_ptr(tensor.buf());

    // Run it
    return run(cmd);
}
```

with:
```cpp
// Put a Tensor on the server
CommandReply Redis::put_tensor(TensorBase& tensor)
{
    // Build the command
    SingleKeyCommand cmd;
    cmd << "AI.TENSORSET" << Keyfield(tensor.name()) << tensor.type_str()
        << tensor.dims() << "BLOB" << tensor.buf();

    // Run it
    return run(cmd);
}
```

# Justification
This will make writing commands faster and more compact. More importantly, it will make reading
and reviewing them much simpler as the constructions will be far more readable.

# Implementation Strategy
Implementation will consist of:
1. Implementing several variants of `operator<<` in `command.h`
2. Implementing a `Keyfield` class that inherits all constructors from `std::strnig` with a `using` statement
3. Updating library and test files to use the new methodology

# Notes
1. While flagging keys by wrapping them in the Keyfield class will work for singleton fields, for vectors of fields (currently added via `add_fields()`), there isn't a good way to mark an entire vector as keys. The `operator<<` method for vectors will thus have to work only for non-key collections, and a new method `add_keys()` for adding vectors of keys.
2. While investigating this refactor, I noticed that several keys are not added as keys in the SmartRedis code.

billschereriii:
* Implement `operator<<()` for command classes
* Implement Keyfield support for loading keys
* Correct missing key denotations in source files

Resolves #222

### DIFF
--- a/src/cpp/redis.cpp
+++ b/src/cpp/redis.cpp
@@ -146,9 +146,8 @@ CommandReply Redis::put_tensor(TensorBase& tensor)
 {
     // Build the command
     SingleKeyCommand cmd;
-    cmd << "AI.TENSORSET" << Keyfield(tensor.name()) << tensor.type_str();
-    cmd.add_fields(tensor.dims());
-    cmd << "BLOB" << tensor.buf();
+    cmd << "AI.TENSORSET" << Keyfield(tensor.name()) << tensor.type_str()
+        << tensor.dims() << "BLOB" << tensor.buf();
 
     // Run it
     return run(cmd);
@@ -213,9 +212,8 @@ CommandReply Redis::copy_tensor(const std::string& src_key,
     // Build a PUT command to send the tensor back to the database
     // under the new key
     MultiKeyCommand cmd_put;
-    cmd_put << "AI.TENSORSET" << Keyfield(dest_key) << TENSOR_STR_MAP.at(type);
-    cmd_put.add_fields(dims);
-    cmd_put << "BLOB" << blob;
+    cmd_put << "AI.TENSORSET" << Keyfield(dest_key) << TENSOR_STR_MAP.at(type)
+            << dims << "BLOB" << blob;
 
     // Run the PUT command
     return run(cmd_put);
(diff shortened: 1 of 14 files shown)


src/cpp/redis.cpp
```diff
@@ -106,8 +106,7 @@ bool Redis::key_exists(const std::string& key)
 {
     // Build the command
     SingleKeyCommand cmd;
-    cmd.add_field("EXISTS");
-    cmd.add_field(key);
+    cmd << "EXISTS" << key;
```
EricGustin, line 109: Should this be `cmd << "EXISTS" << KeyField(key);`?

Spartee: @billschereriii ?

billschereriii: This is murky. The original code did NOT mark this key as a key, and I'm not clear on why. I need to check in with MattE on why to decide whether it should be marked as a key here.

Spartee (commented):
I like it. I think the majority of them look cleaner this way and in general I am a fan of using defined operators for classes whenever possible.

ashao:
Looks good to me: especially as someone whose C++ is a bit rustier, the `<<` makes more sense to me than adding a sequence of fields. The only concern that I have is that it will add some extra code if we start needing to support additional types of fields since we have to duplicate `add_field` and the `<<` operator definition. That being said, the tradeoff is worth it for readability/compactness.

ashao (approved):

src/cpp/redis.cpp
mellis13: We can add this as a ``Keyfield`` if we want.  The ``Redis`` class doesn't need it to be accurate because all requests go to a single node.  I don't think there is a performance impact either way.

PR #233, opened 2022-03-14
Update random number generation to not interfere with client code

billschereriii:
Replace usage of C-style global srand()/rand() methods with C++ local RNGs


ashao (approved):
Looks good to me. Tests on my machine pass just fine. Feel free to merge @billschereriii

PR #229, opened 2022-03-02
Move model commands to new RedisAI interface

billschereriii:
Migrate `model_set ()` and `model_run()` to `AI.MODELSTORE` and `AI.MODELEXECUTE`, respectively; update docs.

Note: `AI.MODELEXECUTE` introduces a timeout for the duration of model execution that was not present in `AI.MODELRUN`. By default, we set a timeout of one hour; users may override this behavior by setting the environment variable `SR_MODEL_TIMEOUT` to a new value (in milliseconds).

This PR partially addresses ticket #133. The `script_set()` and `script_run()` commands have not been updated as this cannot be done without a breaking API change. Therefore ticket #133 should not be closed with this PR.

Issue #133, opened 2021-08-12
Update API to conform with RedisAI changes


mellis13:
# Description
RedisAI has deprecated several API functions and added some additional functionality to the replacement functions.  The SmartRedis API should be updated to conform with these changes.

# Justification
The additional features of the API  are useful for users, and the existing RedisAI API will be removed in future releases.

# Implementation Strategy
The following API changes are needed:

- [ ]  ``ModelSet`` changed to ``ModelStore``.  The ``input_count`` and ``output_count``  fields will need to be added to the command execution.
- [ ]  ``ScriptSet`` changed to ``ScriptStore``.  ``ScriptStore`` has an additional parameter ``ENTRY_POINTS`` which will need to be added to the SmartRedis API.
- [ ]  ``ModelRun`` changed to ``ModelExecute``.  The ``input_count`` and ``output_count``  fields will need to be added to the command execution.
- [ ]  `` ScriptRun`` changed to ``ScriptExecute``.  The additional parameters ``ARGS`` and ``TIMEOUT`` need to be added to the SmartRedis API.  Also, the ``KEYS`` field needs to be added to the command execution.

billschereriii:
Changing `script_set()` to use `AI.SCRIPTSTORE` requires breaking changes to both the SmartRedis API and to user scripts (details below). For this reason, PR #229 only addresses the update for `model_set()` and `model_run()`.

Details: 
1. `AI.SCRIPTSTORE` requires enumeration of script entry points, where `AI.SCRIPTSET` did not. While we could try to parse the user's script to identify these entry points, that is a comparatively large effort that would only serve as a partial solution because of item 2, next:
2. The script entry point functions s for `AI.SCRIPTSTORE`/`AI.SCRIPTEXECUTE` now have a required signature of `def entry_point(tensors: List[Tensor], keys: List[str], args: List[str])`. Users will need to update their scripts to conform to this new interface.

PR #234, opened 2022-03-14
Update RedisAI calls to use non-deprecated API for Models

billschereriii:
Migrate model_set () and model_run() to AI.MODELSTORE and AI.MODELEXECUTE, respectively; update docs.

Note: AI.MODELEXECUTE introduces a timeout for the duration of model execution that was not present in AI.MODELRUN. By default, we set a timeout of one hour; users may override this behavior by setting the environment variable SR_MODEL_TIMEOUT to a new value (in milliseconds).

This PR partially addresses ticket https://github.com/CrayLabs/SmartRedis/issues/133. The script_set() and script_run() commands have not been updated as this cannot be done without a breaking API change. Therefore ticket https://github.com/CrayLabs/SmartRedis/issues/133 should not be closed with this PR.

This is the reopened PR #229

### DIFF
--- a/src/cpp/redis.cpp
+++ b/src/cpp/redis.cpp
@@ -306,7 +306,7 @@ CommandReply Redis::run_model(const std::string& key,
     // Check for a non-default timeout setting
     int run_timeout;
     _init_integer_from_env(run_timeout, _MODEL_TIMEOUT_ENV_VAR,
-                           60 * 60 * 1000); // 1 hr
+                           _DEFAULT_MODEL_TIMEOUT);
 
     // Build the command
     CompoundCommand cmd;
--- a/src/cpp/redis.cpp
+++ b/src/cpp/redis.cpp
@@ -261,7 +261,7 @@ CommandReply Redis::set_model(const std::string& model_name,
 {
     // Build the command
     SingleKeyCommand cmd;
-    cmd << "AI.MODELSET" << Keyfield(model_name) << backend << device;
+    cmd << "AI.MODELSTORE" << Keyfield(model_name) << backend << device;
 
     // Add optional fields if requested
     if (tag.size() > 0) {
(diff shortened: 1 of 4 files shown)


src/cpp/rediscluster.cpp
```diff
@@ -327,10 +327,12 @@ CommandReply RedisCluster::set_model(const std::string& model_name,
     for ( ; node != _db_nodes.cend(); node++) {
         // Build the node prefix
         std::string prefixed_key = "{" + node->prefix + "}." + model_name;
+        std::vector<std::string> tmp_inputs = _get_tmp_names(inputs, node->prefix);
```
mellis13, line 330: Is there are a reason why we are doing _get_tmp_names on these?

src/cpp/redis.cpp
```diff
@@ -303,10 +303,17 @@ CommandReply Redis::run_model(const std::string& key,
                               std::vector<std::string> inputs,
                               std::vector<std::string> outputs)
 {
+    // Check for a non-default timeout setting
+    int run_timeout;
+    _init_integer_from_env(run_timeout, _MODEL_TIMEOUT_ENV_VAR,
+                           60 * 60 * 1000); // 1 hr
```
mellis13, line 309: We we define the default value as a constant like we do for other timeouts?  e.g.:

```c++
static constexpr int _DEFAULT_MODEL_TIMEOUT = 100;
```

src/cpp/rediscluster.cpp
```diff
@@ -389,10 +391,15 @@ CommandReply RedisCluster::set_script(const std::string& key,
...
 // Run a model in the database using the specificed input and output tensors
-CommandReply RedisCluster::run_model(const std::string& key,
+CommandReply RedisCluster::run_model(const std::string& model_name,
                                      std::vector<std::string> inputs,
                                      std::vector<std::string> outputs)
 {
+    // Check for a non-default timeout setting
+    int run_timeout;
+    _init_integer_from_env(run_timeout, _MODEL_TIMEOUT_ENV_VAR,
+                           60 * 60 * 1000); // 1 hr
```
mellis13, line 401: Same comment on using a constant -- just flagging for convenience

mellis13 (approved):
Just a few comments, but should be good to merge with minor revisions.  Thanks!

src/cpp/rediscluster.cpp
billschereriii: This is a bug; fixed

src/cpp/redis.cpp
billschereriii: Good call; fixed

src/cpp/rediscluster.cpp
billschereriii: Good call; fixed

PR #237, opened 2022-03-16
Implement support for chunked models

### LINKED ISSUE #186: Data chunking for models and scripts
# Description
Redis has a hard limit of 512 MB for string values.  This could be a problem for ML models that are greater than 512 MB in size.  
Currently we take in the user model and submit to redis with a single ``BLOB data`` argument.  RedisAI documentation now clarifies that you can submit with ``BLOB chunk1 chunk2 ``.  We should change the SmartRedis ``Command`` construction for ``model_set`` and ``script_set`` to split the model into 512MB chunks.

# Justification
It is likely that users will eventually try to store and execute models larger than 512MB.  We need to make this change to allow them to do so.

# Implementation Strategy

- [ ] Change  SmartRedis ``Command`` construction for ``model_set`` and ``script_set`` to split the model/script into 512MB chunks.

billschereriii:
* Implement model chunking for `set_model()` and `get_model()`
* Add support for AI.CONFIG to set model chunking size
* Update unit tests to verify chunking support
* Remove an extraneous debug print statement

Resolves #186


billschereriii:
This issue is currently blocked by https://github.com/RedisAI/RedisAI/issues/908

PR #239, opened 2022-03-21
Add Redis cluster testing

ben-albrecht:
This PR appends Redis cluster testing to the existing `run_tests` suite, such that the full test suite gets run with a redis cluster across the full test matrix.

The setup involves:

- Installing some additional dependencies
    - docker and redis-server (6.2.5, same version that redisai image uses)
- Installing redisai by copying it out of the docker container into `$HOME`
- Creating a local redis cluster with 3 daemonized redis servers
- Overwriting the `SSDB` and `SMARTREDIS_TEST_CLUSTER` environment variables

Also in this PR:

- updated the redisai service container to use a more recent ubuntu version (bionic, 18.0.4 instead of xenial, 16.0.4) for the image, to keep the images we use consistent.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
+      - redis-cluster-testing
   pull_request:
     branches:
       - master
@@ -16,6 +17,8 @@ env:
   HOMEBREW_NO_BOTTLE_SOURCE_FALLBACK: "ON"
   HOMEBREW_NO_GITHUB_API: "ON"
   HOMEBREW_NO_INSTALL_CLEANUP: "ON"
+  TZ: "America/New_York" # for tzdata apt package
+  DEBIAN_FRONTEND: "noninteractive" # disable interactive apt installs
   SSDB: "127.0.0.1:6379"
   SMARTREDIS_TEST_CLUSTER: False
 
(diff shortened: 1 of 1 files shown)


ben-albrecht:
It turns out `make test-verbose` rebuilds all tests every time even if they have been previously built. It would cut few minutes off each test config if we didn't rebuild tests in the redis cluster run. However, I opted to save that change for a follow-up effort and keep this PR limited to the workflow changes. LMK if you disagree @mellis13.

billschereriii:
Echoing a comment Sam made on the test-examples PR, I think we would be better off with separate run_tests yml files, one for clustered testing and one for non-clustered.

ben-albrecht:
> Echoing a comment Sam made on the test-examples PR, I think we would be better off with separate run_tests yml files, one for clustered testing and one for non-clustered.

@mellis13 and I determined adding it onto the existing test run has the benefits of re-using a lot of the stuff that gets built in a normal run and lets us run redis cluster testing on the full build matrix without adding another 30 configurations. That said, I'm happy to revisit that decision if others feel strongly.

billschereriii:
> > Echoing a comment Sam made on the test-examples PR, I think we would be better off with separate run_tests yml files, one for clustered testing and one for non-clustered.
> 
> @mellis13 and I determined adding it onto the existing test run has the benefits of re-using a lot of the stuff that gets built in a normal run and lets us run redis cluster testing on the full build matrix without adding another 30 configurations. That said, I'm happy to revisit that decision if others feel strongly.

Okay, let's not hold this up just over that. We can always revisit the decision later and split the files if need be.

.github/workflows/run_tests.yml
```diff
@@ -116,3 +117,40 @@ jobs:
...
+        run: |
+          sudo apt-get update && sudo apt-get -y install curl gnupg lsb-release software-properties-common ca-certificates && \
+          # Add latest redis to apt sources
+          echo "deb [signed-by=/usr/share/keyrings/redis-archive-keyring.gpg] https://packages.redis.io/deb $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/redis.list && \
+          curl -fsSL https://packages.redis.io/gpg | gpg --dearmor -o /usr/share/keyrings/redis-archive-keyring.gpg && \
+          # Add latest docker to apt sources
+          echo  "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null && \
+          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg && \
+          sudo apt-get update && \
+          sudo apt-get -y install iputils-ping docker-ce docker-ce-cli containerd.io redis-server
```
mellis13, line 131: This might be out of the scope of this PR, but just wanted to kick off a discussion of whether we should be using a specific version of redis in our tests so we can more intentionally change version numbers to match SmartSim.  I see advantages to both options of pinning to a version or just using latest.  I think SmartSim uses 6.0.8 and I think this is pulling down 6.2.6.  This doesn't need to hold up this PR, we can create a follow on issue if we decide to match the smartsim version.

mellis13 (approved):
This is a great improvement to the testing actions.  Thank you!

.github/workflows/run_tests.yml
mellis13: After thinking more about this, it might make more sense to make the redis-server version identical to the one used in the redisai docker image so that the non-cluster and cluster images are more consistent.

PR #240, opened 2022-03-22
delete_model and delete_script

billschereriii:
Implement delete_model and delete script for C, C++, Fortran, Python clients; update unit tests to add coverage of same


al-rigazzi (approved):
Thanks for implementing these calls, they will be very useful in some workflows.

LGTM!

al-rigazzi:
@billschereriii  Just a question, do we know why one CI action failed?

tests/fortran/client_test_ensemble.F90
```diff
@@ -142,5 +142,22 @@ program main
   error stop
 endif
 
+result = client%delete_model(model_key)
+if (result .ne. SRNoError) stop
```
ashao, line 146: These should be `error stop` otherwise these errors won't register as a failed test. There's a bigger chore to change that elsewhere within the Fortran tests, but for now would you mind updating them in `client_test_ensemble.F90` for now?

ashao (changes requested):
One minor comment about `stop` vs `error stop`. Otherwise this looks good to me. I'll test on WSL-Ubuntu while you're updating the PR.

src/cpp/rediscluster.cpp
```diff
@@ -505,6 +505,54 @@ CommandReply RedisCluster::run_script(const std::string& key,
...
+// Delete a script from the database
+CommandReply RedisCluster::delete_script(const std::string& key)
+{
+    CommandReply reply;
+    std::vector<DBNode>::const_iterator node = _db_nodes.cbegin();
+    for ( ; node != _db_nodes.cend(); node++) {
+        // Build the node prefix
+        std::string prefixed_key = "{" + node->prefix + "}." + key;
+
+        // Build the SCRIPTSET command
```
mellis13, line 541: I think this comment needs updated

src/cpp/rediscluster.cpp
```diff
@@ -505,6 +505,54 @@ CommandReply RedisCluster::run_script(const std::string& key,
...
+// Delete a model from the database
+CommandReply RedisCluster::delete_model(const std::string& key)
+{
+    CommandReply reply;
+    std::vector<DBNode>::const_iterator node = _db_nodes.cbegin();
+    for ( ; node != _db_nodes.cend(); node++) {
+        // Build the node prefix
+        std::string prefixed_key = "{" + node->prefix + "}." + key;
+
+        // Build the SCRIPTSET command
```
mellis13, line 517: I think this comment needs updated

ashao (approved):

PR #246, opened 2022-04-13
AddressAllCommand

billschereriii:
Implement AddressAllCommand and refactor existing commands to use it.
Note: No need for additional test cases as the existing ones cause all the functionality to be exercised


src/cpp/rediscluster.cpp
```diff
@@ -126,6 +126,43 @@ CommandReply RedisCluster::run(AddressAnyCommand &cmd)
     return _run(cmd, _last_prefix);
 }
 
+// Run a non-keyed Command that addresses any db node on the server
```
ashao, line 129: Should 'server' be 'cluster'?

src/cpp/rediscluster.cpp
```diff
@@ -126,6 +126,43 @@ CommandReply RedisCluster::run(AddressAnyCommand &cmd)
...
+    if (cmd.key_index != -1 && cmd.get_field_count() < cmd.key_index) {
+        throw SRInternalException("Invalid key_index executing command!");
+    }
+
+    // Find the segment to be tweaked for each node
+    std::string field;
+    if (cmd.key_index != -1) {
+        Command::const_iterator it = cmd.cbegin();
+        int i = 0;
+        for ( ; it != cmd.cend(); it++) {
```
ashao, line 142: Coming from a place of relatively little knowledge, do we only expect one field to be set in a given command? Otherwise this loop as written seems like it would just return the last one even if it detects more than one.

src/cpp/rediscluster.cpp
```diff
@@ -322,39 +359,34 @@ CommandReply RedisCluster::set_model(const std::string& model_name,
...
-            cmd << "MINBATCHSIZE" << std::to_string(min_batch_size);
-        }
-        if ( inputs.size() > 0) {
-            cmd << "INPUTS" << std::to_string(inputs.size()) << inputs;
-        }
-        if (outputs.size() > 0) {
-            cmd << "OUTPUTS" << std::to_string(outputs.size()) << outputs;
-        }
-        cmd << "BLOB" << model;
+    // Add optional fields as requested
```
ashao, line 368: Are the optional arguments order dependent at all? Also, gotta say it makes it so nice to build the command via `<<`

src/cpp/command.cpp
```diff
@@ -119,6 +119,44 @@ void Command::add_field(std::string field, bool is_key)
...
+//  Replace a field in a command
+void Command::set_field_at(std::string new_field, size_t pos, bool is_key)
+{
+    size_t field_size = new_field.size();
+    char* f = NULL;
+    try {
+        f = (char*)new unsigned char[field_size + 1];
+    }
+    catch (std::bad_alloc& e) {
+        throw SRBadAllocException("field");
```
ashao, line 131: Make this a little more descriptive.

src/cpp/rediscluster.cpp
```diff
@@ -126,6 +126,43 @@ CommandReply RedisCluster::run(AddressAnyCommand &cmd)
...
+    for ( ; node != _db_nodes.cend(); node++) {
+        // swap in a replacement segment for each one
+        std::string new_field = "{" + node->prefix + "}." + field;
+        cmd.set_field_at(new_field, cmd.key_index, true);
+
+        // Execute the updated command
+        cmd.set_exec_address_port(node->ip, node->port);
+        reply = _run(cmd, node->prefix);
+        if (reply.has_error() > 0)
+            break; // Short-circuit failure on error
```
ashao, line 161: Why break and not just throw the error here?

ashao (changes requested):
Looks pretty good to me. I have some minor critiques that might be due to a lack of in-depth knowledge about the the structure of Redis commands, fields, and keys are handled. However, if you can explain why those questions/critiques are not potential problems, I'd appreciate knowing that.

src/cpp/rediscluster.cpp
billschereriii: We use the terms interchangeably throughout this file

src/cpp/rediscluster.cpp
billschereriii: Yes, there should only be a single key

src/cpp/rediscluster.cpp
billschereriii: Yes, the order is prescribed

src/cpp/command.cpp
billschereriii: This is our standard for memory allocation failures.

src/cpp/command.cpp
```diff
@@ -119,6 +119,44 @@ void Command::add_field(std::string field, bool is_key)
...
+    }
+
+    new_field.copy(f, field_size, 0);
+    f[field_size] = '\0';
+
+    // Free memory from the previous occupant of this
```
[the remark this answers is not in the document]
billschereriii: Local memory allocation happens when the field was originally submitted from a std::string -- we make a copy of the contents in an allocated memory block

src/cpp/rediscluster.cpp
billschereriii: Largely stylistic. I wanted to keep as much extraneous stuff out of the loop as possible to make it easy to read what normal (successful execution) path is doing.  With just a single return statement, yes, I could have done that, but if we ever go back and add more stuff in the case of a failure, it will be more complex.

ashao (approved):
Thanks for taking the time to walk me through the implementation

PR #248, opened 2022-04-14
Data aggregation C++ API

mellis13:
This is a draft PR for the data aggregation API in C++.  A draft PR has been put up so parallel work can be done multi-threading the dataset retrieval from cluster shards.  CI tests will likely fail until all functionality is done.

The following items still need to be addressed before final review and merge of this PR:

- [x] Implement ``rename_list()``
- [x] Tests for ``rename_list()``
- [x] Implement ``copy_list()``
- [x] Tests for ``copy_list()``
- [x] Implement ``Redis::run_via_unordered_pipelines()``.  This was implemented in ``RedisCluster`` and the ``Redis`` version of this function will be extremely easy since all keys are colocated on one shard.
- [x] More tests to cover error statements to ensure code coverage.
- [x] Cleanup ``Client::_get_dataset_list_range`` to functionalize the code for reuse in other areas (e.g. ``Client.get_dataset()``
- [x] Resolve whether we want the retrieved dataset names to be the dataset key or the actual names provided by the user (discussion to follow).
- [x] Nest pipeline execution so that redis-plus-plus errors can be caught
- [x] Documentation on the data aggregation API
- [x] Threading on the innermost loop (i.e. ``RedisCluster::run_via_unordered_pipelines()``).  This will be evaluated in a separate ticket, and this PR should not be merged before performance data is available from that.

### DIFF
--- a/doc/data_structures.rst
+++ b/doc/data_structures.rst
@@ -366,8 +366,11 @@ function to retrieve the ``aggregation list`` contents is shown below:
 
 Additional functions are provided to retrieve only a portion of the
 ``aggregation list`` contents, copy an ``aggregation list``, rename
-an ``aggregation list``, retrieve ``aggregation list`` length, and
-poll for a specified ``aggregation list`` length.
+an ``aggregation list` and retrieve ``aggregation list`` length.
+A blocking method to poll the ``aggregation list`` length is also
+provided as a means to wait for list completion before performing
+another task in the same application or a separate application.
+
 
 Model
 =====
--- a/doc/data_structures.rst
+++ b/doc/data_structures.rst
@@ -324,6 +324,51 @@ metadata field to be retrieved, and this name is the same name that
 was used when constructing the metadata field with
 ``add_meta_scalar()`` and ``add_meta_string()`` functions.
 
+Aggregating
+-----------
+
+An API is provided to aggregate multiple ``DataSet`` objects that
+are stored on one or more database nodes.  This is accomplished
+through an interface referred to as ``aggregation lists``.
+An ``aggregation list`` in SmartRedis stores references to
+``DataSet`` objects that are stored in the database.  ``DataSet``
+objects can be appended to the ``aggregation list`` and then
+``SmartRedis`` clients in the same application or a different application
+can retrieve all or some of the ``DataSet`` objects referenced in that
+``aggregation list``.
+
+For example, the C++ client function to append a ``DataSet`` to an
+aggregation list is shown below:
+
+.. code-block:: cpp
+
+    # C++ aggregation list append interface
+    void append_to_list(const std::string& list_name,
+                        const DataSet& dataset);
+
+The above function will append the provided ``DataSet`` to the
+``aggregation list``, which can be referenced in all user-facing functions
+by the provided list name.  Note that a list can be appended by
+any client in the same or different application.  Additionally, all
+appends are performed at the end of the list, and if the list does not
+already exist, it is automatically created.
+
+For retrieval of ``aggregation list`` contents,
+the SmartRedis ``Client`` method provides an API function that
+will return an iterable container with all of the ``DataSet`` objects
+that were appended to the ``aggregation list``.  For example, the C++ client
+function to retrieve the ``aggregation list`` contents is shown below:
+
+.. code-block:: cpp
+
+    # C++ aggregation list retrieval interface
+    std::vector<DataSet> get_datasets_from_list(const std::string& list_name);
+
+Additional functions are provided to retrieve only a portion of the
+``aggregation list`` contents, copy an ``aggregation list``, rename
+an ``aggregation list``, retrieve ``aggregation list`` length, and
+poll for a specified ``aggregation list`` length.
+
 Model
 =====
 
--- a/include/client.h
+++ b/include/client.h
@@ -1123,6 +1123,7 @@ class Client
         *            name. See set_data_source() and use_list_ensemble_prefix()
         *            for more details.
         *   \param name The name of the list
+        *   \param list_length The desired length of the list
         *   \param poll_frequency_ms The time delay between checks,
         *                            in milliseconds
         *   \param num_tries The total number of times to check for the name
@@ -1130,7 +1131,7 @@ class Client
         *            than or equal to the provided length, otherwise false
         *   \throw SmartRedis::Exception if poll list length command fails
         */
-        bool poll_list_length(const std::string& name, size_t list_length,
+        bool poll_list_length(const std::string& name, int list_length,
                               int poll_frequency_ms, int num_tries);
 
         /*!
--- a/include/client.h
+++ b/include/client.h
@@ -1544,6 +1544,20 @@ class Client
         */
         void _unpack_dataset_metadata(DataSet& dataset,
                                       CommandReply& reply);
+
+        /*!
+        *   \brief Determine the dataset name from a aggregation
+        *          list entry
+        *   \details DataSet aggregation lists store the key
+        *            of the DataSet so that metadata and tensor
+        *            keys can be retrieved using the entry.  The list
+        *            entry is of the form ensemble_member.{dataset_name}.
+        *   \param dataset_key The dataset key in the form of
+        *                      ensemble_member.{dataset_name}
+        *   \returns DataSet name
+        */
+        std::string _get_dataset_name_from_list_entry(std::string& dataset_key);
+
 };
 
 } //namespace SmartRedis
--- a/include/client.h
+++ b/include/client.h
@@ -1609,7 +1609,7 @@ class Client
         *   \param poll_frequency_ms The time delay between checks,
         *                            in milliseconds
         *   \param num_tries The total number of times to check for the name
-        *   \param comparator The comparison function
+        *   \param comp_func The comparison function
         *   \returns Returns true if the list is found with a length greater
         *            than or equal to the provided length, otherwise false
         *   \throw SmartRedis::Exception if poll list length command fails
--- a/include/client.h
+++ b/include/client.h
@@ -759,7 +759,7 @@ class Client
 
         /*!
         *   \brief Check if a key exists in the database, repeating the check
-        *          the check at a specified polling interval
+        *          at a specified polling interval
         *   \param key The key to be checked in the database
         *   \param poll_frequency_ms The time delay between checks,
         *                            in milliseconds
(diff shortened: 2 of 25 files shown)


include/client.h
```diff
@@ -759,7 +759,7 @@ class Client
 
         /*!
         *   \brief Check if a key exists in the database, repeating the check
-*       *          the check at a specified polling interval
+        *          the check at a specified polling interval
```
billschereriii, line 762: "the check" is written twice here

include/client.h
```diff
@@ -1026,6 +1051,132 @@ class Client
...
+        /*!
+        *   \brief Appends a dataset to the aggregation list
+        *   \details When appending a dataset to an aggregation list,
+        *            the list will automatically be created if it does not
+        *            exist (i.e. this is the first entry in the list).
+        *            Aggregation lists work by referencing the dataset
+        *            by storing its key, so appending a dataset
+        *            to an aggregation list does not create a copy of the
+        *            dataset.  Also, for this reason, the dataset
+        *            must be placed into the database with a separate call
```
billschereriii, line 1063: Must have _previously_ been placed into the dataset?

include/client.h
```diff
@@ -1026,6 +1051,132 @@ class Client
...
+        *            and use_list_ensemble_prefix() for more details.
+        *   \param old_name The old list name
+        *   \param new_name The new list name
+        *   \throw SmartRedis::Exception if the command fails
+        */
+        void rename_list(const std::string& src_name,
+                         const std::string& dest_name);
+
+        /*!
+        *   \brief Get the length of an aggregation list
```
billschereriii, line 1111: Rather than repeat the name of the function, how about something like "Get the number of elements in an aggregation list"

include/client.h
```diff
@@ -1026,6 +1051,132 @@ class Client
...
+        */
+        std::vector<DataSet> get_datasets_from_list(const std::string& list_name);
+
+        /*!
+        *   \brief Get a range of datasets (by index) from an aggregation list
+        *   \details The aggregation list key used to retrieve datasets
+        *            may be formed by applying a prefix to the supplied
+        *            name. See set_data_source()  and use_list_ensemble_prefix()
+        *            for more details.  An empty or nonexistant aggregation
+        *            list returns an empty vector.  If the provided
```
billschereriii, line 1156: Nonexistent list should throw a runtime exception

include/client.h
```diff
@@ -1026,6 +1051,132 @@ class Client
...
+        *                    supported.  A negative value indicates offsets
+        *                    starting at the end of the list. For example, -1 is
+        *                    the last element of the list.
+        *   \returns A vector containing DataSet objects.
+        *   \throw SmartRedis::Exception if retrieval fails or
+        *          input parameters are invalid
+        */
+        std::vector<DataSet> get_dataset_list_range(const std::string& list_name,
+                                                    const int start_index,
+                                                    const int end_index);
```
billschereriii, line 1178: Rather than make the user calculate the range, should we mirror the approach in the multiGPU functions and have them provide the total number of processes plus their current rank -- and then provide the relevant "slice" of data?

src/cpp/client.cpp
```diff
@@ -1090,6 +1101,222 @@ void Client::save(std::string address)
         throw SRRuntimeException("SAVE command failed");
 }
 
+void Client::append_to_list(const std::string& list_name,
```
billschereriii, line 1104: Add one-liners before functions to very briefly describe what they do

src/cpp/client.cpp
```diff
@@ -1090,6 +1101,222 @@ void Client::save(std::string address)
...
+        }
+
+        // Check that the string length is not zero
+        if(reply[i].str_len() == 0) {
+            throw SRRuntimeException("Element " + std::to_string(i) +
+                                     " contains an empty key, which is "\
+                                     "not permitted.");
+        }
+        std::cout<<"Adding "<<std::string(reply[i].str(), reply[i].str_len())<<std::endl;
+        copy_cmd.add_field_ptr(reply[i].str(), reply[i].str_len());
```
billschereriii, line 1215: Consider rewriting as:
copy_cmd << std::string(reply[i].str(), reply[i].str_len());

src/cpp/client.cpp
```diff
@@ -1169,6 +1396,167 @@ inline CommandReply Client::_get_dataset_metadata(const std::string& name)
...
+                                     "not permitted.");
+        }
+
+        // Get the dataset key from the list entry
+        std::string dataset_key_prefix(reply[i].str(), reply[i].str_len());
+
+        // Build the metadata retrieval command
+        SingleKeyCommand* metadata_cmd =
+            metadata_cmd_list.add_command<SingleKeyCommand>();
+        (*metadata_cmd) << "HGETALL" << Keyfield(dataset_key_prefix + ".meta");
```
billschereriii, line 1468: Can we make a _build_metadata_key() routine so we don't have ".meta" embedded into functions like this? It'll make maintenance a lot easier later.

src/cpp/pipelinereply.cpp
```diff
@@ -0,0 +1,109 @@
...
+size_t PipelineReply::size()
+{
+    return _all_replies.size();
+}
+
+// Check if any CommandReply has error
+bool PipelineReply::has_error()
+{
+    for (size_t i = 0; i < _all_replies.size(); i++) {
+        if (CommandReply::shallow_clone(_all_replies[i]).has_error() > 0)
```
billschereriii, line 79: Seems wasteful to make a clone when all we need to do is read a field

tests/cpp/unit-tests/test_aggregation_list.cpp
```diff
@@ -0,0 +1,438 @@
...
+                    int list_length = client.get_list_length(list_name);
+                    CHECK((size_t)list_length == datasets.size());
+                }
+
+                AND_THEN("Polling for the correct list length exits "\
+                         "immediately")
+                {
+                    int list_length = datasets.size();
+                    CHECK(client.poll_list_length(list_name,
+                                                  list_length, 5, 100));
```
billschereriii, line 187: should that 5 be a 1?

include/client.h
```diff
@@ -1026,6 +1051,132 @@ class Client
...
+        *          attemps is exceeded, false is returned.
+        *   \details The aggregation list key used to check for list length
+        *            may be formed by applying a prefix to the supplied
+        *            name. See set_data_source() and use_list_ensemble_prefix()
+        *            for more details.
+        *   \param name The name of the list
+        *   \param list_length The desired length of the list
+        *   \param poll_frequency_ms The time delay between checks,
+        *                            in milliseconds
+        *   \param num_tries The total number of times to check for the name
```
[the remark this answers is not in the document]
mellis13: No.  None of the ``poll`` functions in SmartRedis support -1 for infinity.

include/client.h
mellis13: The group thought that throwing a runtime exception because the list is temporarily empty will make it hard to distinguish between an empty list (which might be useful when dividing work among processes for training) and an execution error.  With future workflows in mind, the group decided to align with the Redis standard of returning an empty vector.

include/client.h
```diff
@@ -1026,6 +1051,132 @@ class Client
...
+        *            for more details.  An empty or nonexistant
+        *            aggregation list returns an empty vector.
+        *   \param list_name The name of the aggregation list
+        *   \returns A vector containing DataSet objects.
+        *   \throw SmartRedis::Exception if retrieval fails.
+        */
+        std::vector<DataSet> get_datasets_from_list(const std::string& list_name);
+
+        /*!
+        *   \brief Get a range of datasets (by index) from an aggregation list
```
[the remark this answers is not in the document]
mellis13: We can think about this for future work.  It might be helpful shorthand for debugging. As it is, the partial keys stored in the ``aggregation list`` couldn't be fed into any other functions because of prefixing constraints, but maybe it is good for diagnostics.

src/cpp/client.cpp
mellis13: I started doing this, but the ``_build`` routines we use for keys are actually incompatible with the aggregation list, and it requires a lot of code changes merge the two key construction procedures.  I made a note when doing that work to create an issue to rework how we do the key construction routines to make it easier.

tests/cpp/unit-tests/test_aggregation_list.cpp
```diff
@@ -0,0 +1,438 @@
...
+                         "to an empty string")
+                {
+                    THEN("A SmartRedis::ParameterException is thrown")
+                    {
+                        CHECK_THROWS_AS(client.rename_list(list_name, ""),
+                                        ParameterException);
+                    }
+                }
+
+                AND_WHEN("The aggregation list is renamed to the same name")
```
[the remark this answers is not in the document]
mellis13: I think it already exists at line 283?

src/cpp/client.cpp
billschereriii: I'm good with that approach.  We definitely need some structure around key generation. I think it could be a good first issue for someone

src/cpp/pipelinereply.cpp
mellis13: The ``_all_replies`` variable stores ``redisReply*`` types.  To re-use the recursive error checking code in CommandReply, I opted for the shallow copy (which is basically one pointer set to NULL and one pointer copy) over re-rewriting error checking code.  I'm open to rewriting the error checking code if we think it's a big perf hit.

tests/cpp/unit-tests/test_aggregation_list.cpp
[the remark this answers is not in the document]
billschereriii: Indeed, my bad

src/cpp/client.cpp
```diff
@@ -1090,6 +1101,227 @@ void Client::save(std::string address)
...
+                                                    const int end_index)
+{
+    if (list_name.size() == 0) {
+        throw SRParameterException("The list name must have length "\
+                                   "greater than zero");
+    }
+
+    return _get_dataset_list_range(list_name, start_index, end_index);
+}
+
```
ashao, line 1324: A `pop_dataset_list` could be a useful feature to have if you have multiple consumers of an aggregated list.

ashao (changes requested):
Really nice work here @mellis13. You and @billschereriii already anticipated/addressed the things that I thought might have been issues. The main change that I'm suggesting is to add/expose `pop` functionality to the list in the event that you have multiple consumers of the aggregated list.

doc/data_structures.rst
```diff
@@ -324,6 +324,51 @@ metadata field to be retrieved, and this name is the same name that
...
+.. code-block:: cpp
+
+    # C++ aggregation list retrieval interface
+    std::vector<DataSet> get_datasets_from_list(const std::string& list_name);
+
+Additional functions are provided to retrieve only a portion of the
+``aggregation list`` contents, copy an ``aggregation list``, rename
+an ``aggregation list``, retrieve ``aggregation list`` length, and
+poll for a specified ``aggregation list`` length.
+
```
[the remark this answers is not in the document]
mellis13: Good idea -- I've updated the docs.

ashao (approved):
Changes all look good to me

PR #250, opened 2022-04-25
multiGPU delete for scripts and models

### LINKED ISSUE #247: model_delete_multigpu() and script_delete_multigpu()
# Description
Multi-GPU variants of `delete_model()` and `delete_script()` are needed to properly delete models and scripts stored via `set_model_multigpu()` and `set_script_multigpu()`.

# Justification
The existing `delete_model()` and `delete_script()` methods will not clean up the copies of models and scripts that target specific GPUs.

# Implementation Strategy
In both cases, a simple loop through GPUs will  give the names of models and scripts to be deleted via the existing routines. A test case can be added that will set the models and scripts for multi-GPU, then delete them. This will improve code coverage without requiring access to a GPU system (as long as we don't try running the model/script).

billschereriii:
Implement `delete_model_multigpu()` and `delete_script_multigpu()` for C++ and Python; update user documentation to detail these new functions; update test scripts to test them (as much as possible).

Resolves #247


src/cpp/redis.cpp
```diff
@@ -416,19 +416,8 @@ CommandReply Redis::run_script(const std::string& key,
...
-*   \param inputs The names of input tensors to use in the script
-*   \param outputs The names of output tensors that will be used
-*                  to save script results
-*   \param offset index of the current image, such as a processor
-*                   ID or MPI rank
-*   \param num_gpus the number of gpus for which the script was stored
-*   \throw RuntimeException for all client errors
-*/
+// Run a script function in the database using the
+// specificed input and output tensors in a multi-GPU system
```
ashao, line 420: 'specificed' to 'specified'

ashao: Another minor editorial comment: "Run a script function in the database with multiple GPUs using the specified input and output tensors"

ashao (approved):
Other than that one editorial change, this looks good to me. Thanks for doing that so quickly.

PR #251, opened 2022-04-28
Multi-threaded pipeline execution

billschereriii:
Implement a thread pool for multi-threaded execution of jobs.
Refactor pipeline execution into a series of jobs dispatched to worker threads via the thread pool.


include/redisserver.h
```diff
@@ -596,6 +614,12 @@ class RedisServer {
         inline static const std::string _MODEL_TIMEOUT_ENV_VAR =
             "SR_MODEL_TIMEOUT";
 
+        /*!
+        *   \brief Environment variable for thread count in thread pool
+        */
+        inline static const std::string _TP_THREAD_COUNT =
+            "SR_THREAD_COUNT";
```
ashao, line 621: Out of scope for this PR, but at some point we should document what environment variables the client initialization depends on. I vaguely wonder if we should create specific methods to initialize these internal variables from the environment or some other method so that we can be flexible on how the clients can be configured.

include/threadpool.h
```diff
@@ -0,0 +1,86 @@
...
+*/
+class ThreadPool
+{
+  public:
+    /*!
+    *   \brief ThreadPool constructor
+    *   \param num_threads The number of threads to create in the pool,
+    *          or 0 to use one thread per hardware context
+    */
+    ThreadPool(unsigned int num_threads=0);
```
ashao, line 24: What about changing the default to `1`? That would allow `num_threads=0` to be a special case where we should bypass creating a threadpool.

src/cpp/rediscluster.cpp
```diff
@@ -224,28 +224,75 @@ PipelineReply RedisCluster::run_via_unordered_pipelines(CommandList& cmd_list)
...
+                    all_replies += std::move(reply);
+                }
+
+                // Increment completion counter
+                ++pipeline_completion_count;
+            }  // Release the lock
+        });
+    }
 
+    // Wait until all jobs have finished
```
ashao, line 287: Should there be a sleep here so we're not slamming this loop continually?

src/cpp/redisserver.cpp
```diff
@@ -52,8 +54,19 @@ RedisServer::RedisServer()
 
     _command_attempts = (_command_timeout * 1000) /
                          _command_interval + 1;
+
+    _tp = new ThreadPool(_thread_count);
```
ashao, line 58: Following up from a previous comment, is there any overhead associated with creating a ThreadPool? I worry a little bit that client initialization is beginning to take a long time. While the cost is amortized over the length of the run, it's still something we need to keep an eye on.

src/cpp/threadpool.cpp
```diff
@@ -0,0 +1,164 @@
...
+using namespace SmartRedis;
+using namespace std::chrono_literals;
+
+// TIME_THREADPOOL
+// Enable this flag to display timings for the threadpool activity.
+// Currently, this will be to screen, but eventually it will go to
+// the SmartRedis log
+#undef TIME_THREADPOOL
+
+// C-tor
```
ashao, line 20: Replace "C-tor" -> Constructor. (Flashbacks to F77 codebases)

src/cpp/threadpool.cpp
```diff
@@ -0,0 +1,164 @@
...
+        threads.push_back(std::thread(&ThreadPool::perform_jobs, this, i));
+    }
+
+    // Announce that we're open for business
+    shutting_down = false;
+    shutdown_complete = false;
+    initialization_complete = true;
+}
+
+// D-tor
```
ashao, line 45: D-tor -> Destructor

src/cpp/threadpool.cpp
```diff
@@ -0,0 +1,164 @@
...
+    if (!shutdown_complete) {
+        shutdown();
+    }
+}
+
+// Shut down the thread pool, blocking any further jobs from submission
+void ThreadPool::shutdown()
+{
+    // Make sure initialization is complete before we shut down
+    while (!initialization_complete)
```
ashao, line 58: Here also, should there be a small wait to avoid hammering the loop?

src/cpp/threadpool.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include "srexception.h"
+
+using namespace SmartRedis;
+using namespace std::chrono_literals;
+
+// TIME_THREADPOOL
+// Enable this flag to display timings for the threadpool activity.
+// Currently, this will be to screen, but eventually it will go to
+// the SmartRedis log
+#undef TIME_THREADPOOL
```
ashao, line 18: Are we controlling behavior via CPP flags elsewhere? As much as it makes me uncomfortable, perhaps this could also be a different set of environment variables.

ashao (changes requested):
Nice work on a good clean implementation of the multi-threading! A couple of minor comments related to the spin loops when trying to ensure all threads are complete.

src/cpp/threadpool.cpp
billschereriii: This is never something that an end user would have any reason to do, just a developer. When we get logging implemented and add a debug flag, this code can be switched to be dependent on the debug flag, but until then, this keeps the code around without interfering with production usage.

src/cpp/threadpool.cpp
billschereriii: If we're shutting down, then there's nothing else we care about doing, so the spin-wait is appropriate

include/threadpool.h
billschereriii: For performance, we want to use as many threads as possible without inducing context switches.  The default, 0, does this.  I don't expect users to be manually setting the environment variable for thread count in normal circumstances

include/redisserver.h
billschereriii: Agreed, let's make sure to catch this variable when we do the variables documentation prior to the release

src/cpp/rediscluster.cpp
ben-albrecht: Would this spin-waiting become an issue for a sufficiently large threadpool with a lot high level of contention? Curious what scale is more representative of our typical workloads.

src/cpp/redisserver.cpp
ben-albrecht: Do we have any performance tests measuring the initialization time, so that we can keep track of this over time?

src/cpp/threadpool.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+// Worker thread main loop to acquire and perform jobs
+void ThreadPool::perform_jobs(unsigned int tid)
+{
+    #ifdef TIME_THREADPOOL
+    int jobid = 0;
+    std::cout << "Thread " << std::to_string(tid) << " reporting for duty" << std::endl;
+    #endif
+
+    // Loop foorever processing jobs until we get killed
```
ben-albrecht, line 97:
```suggestion
    // Loop forever processing jobs until we get killed
```

src/cpp/redisserver.cpp
ben-albrecht: If not, maybe we can introduce a `#TIME_THREADPOOL_INIT` segment to enable that in the future?

ben-albrecht (approved):
Looks great @billschereriii, nice work! I only had a few questions and minor comments within.

src/cpp/rediscluster.cpp
billschereriii: Seems unlikely to me, but if it ever becomes a problem we can go back and add spin-then-wait functionality here. Unless it does I'd rather avoid the complexity.

src/cpp/redisserver.cpp
billschereriii: The overhead wasn't noticeable when I watched the tests running on my Mac. I would guess it's on the order of a millisecond or two right now, not nearly enough to worry about

billschereriii: The overhead wasn't noticeable when I watched the tests running on my Mac. I would guess it's on the order of a millisecond or two right now, not nearly enough to worry about

ashao (approved):
Looks good to me. Please rerun the tests prior to merge to verify that it was a false negative.

PR #254, opened 2022-05-12
Implement multi-GPU C/Fortran methods

ashao:
Convenience functions for multiple GPU systems had previously been
implemented in C++. This adds support for C and Fortran clients
for `[set,delete,run]_[model,script]_multigpu`. This also adds a variant
of the existing MNIST test that uses these new methods. The running
of this test can be toggled by setting the new environment variable
`SMARTREDIS_TEST_DEVICE=gpu`. This is by default set (noisly) in
`setup_test_env.sh` to CPU. Tests pass on `osprey`.


src/c/c_client.cpp
```diff
@@ -459,6 +459,52 @@ bool CompareCaseInsensitive(const char* a,const char* b) {
...
+
+  // For the inputs and outputs arrays, a single empty string is ok (this means
+  // that the array should be skipped) but if more than one entry is present, the
+  // strings must be nonzero length
+  if (_isTensorFlow(backend)) {
+    if (n_inputs != 1 && input_lengths[0] != 0) {
+      for (size_t i = 0; i < n_inputs; i++){
+        if (inputs[i] == NULL || input_lengths[i] == 0) {
+          throw SRParameterException(
+            std::string("inputs[") + std::to_string(i) + "] is NULL or empty");
```
billschereriii, line 493: Because you're adding to a std::string (of `i`), I don't think you need to invoke the `std::string` constructor for "inputs["

src/c/c_client.cpp
```diff
@@ -459,6 +459,52 @@ bool CompareCaseInsensitive(const char* a,const char* b) {
...
+        }
+      }
+    }
+    if (n_outputs != 1 && output_lengths[0] != 0) {
+      for (size_t i = 0; i < n_outputs; i++) {
+        if (outputs[i] == NULL || output_lengths[i] == 0) {
+          throw SRParameterException(
+            std::string("outputs[") + std::to_string(i) + "] is NULL or empty");
```
billschereriii, line 501: Because you're adding to a std::string (of `i`), I don't think you need to invoke the `std::string` constructor for "outputs["

src/c/c_client.cpp
```diff
@@ -476,50 +522,83 @@ SRError set_model_from_file(void* c_client,
   SRError result = SRNoError;
   try
   {
+    _check_params_set_model(c_client, name, backend, inputs, input_lengths, n_inputs,
```
billschereriii, line 525: Move this call below the comment

src/c/c_client.cpp
```diff
@@ -476,50 +522,83 @@ SRError set_model_from_file(void* c_client,
...
+                                     const int batch_size, const int min_batch_size,
+                                     const char* tag, const size_t tag_length,
+                                     const char** inputs, const size_t* input_lengths,
+                                     const size_t n_inputs, const char** outputs,
+                                     const size_t* output_lengths, const size_t n_outputs)
+{
+  SRError result = SRNoError;
+  try
+  {
+    _check_params_set_model(c_client, name, backend, inputs, input_lengths, n_inputs,
```
billschereriii, line 587: Move call under comment

src/c/c_client.cpp
```diff
@@ -679,6 +795,41 @@ SRError get_model(void* c_client,
   return result;
 }
 
+// Put a script in the database that is stored in a file in a multi-GPU system
```
billschereriii, line 798: Please be consistent in ordering all the multiGPU functions after their regular  (non-GPU) versions.  Currently, there's a mix with some before and some after

src/fortran/client.F90
```diff
@@ -677,6 +693,102 @@ function set_model_from_file(self, name, model_file, backend, device, batch_size
...
+                             c_backend, backend_length, c_first_gpu, c_num_gpus, c_batch_size, c_min_batch_size, &
+                             c_tag, tag_length, inputs_ptr, input_lengths_ptr, n_inputs, outputs_ptr, &
+                             output_lengths_ptr, n_outputs)
+  deallocate(c_inputs)
+  deallocate(input_lengths)
+  deallocate(ptrs_to_inputs)
+  deallocate(c_outputs)
+  deallocate(output_lengths)
+  deallocate(ptrs_to_outputs)
+  deallocate(c_tag)
```
billschereriii, line 789: Is it safe to unconditionally deallocate memory that was conditionally allocate?

src/fortran/client.F90
```diff
@@ -881,6 +1162,36 @@ function set_script(self, name, device, script) result(code)
...
+  ! Local variables
+  character(kind=c_char, len=len_trim(name)) :: c_name
+  character(kind=c_char, len=len_trim(script)) :: c_script
+
+  integer(kind=c_size_t) :: name_length
+  integer(kind=c_size_t) :: script_length
+  integer(kind=c_size_t) :: device_length
+  integer(kind=c_int)    :: c_first_gpu, c_num_gpus
+
+  c_name    = trim(name)
```
billschereriii, line 1183: Extra space if you're trying to line up with the next line

src/fortran/client/script_interfaces.inc
```diff
@@ -73,6 +90,24 @@ interface
...
+interface
+  function set_script_multigpu_c( c_client, key, key_length, script, script_length, & 
+                                  first_gpu, num_gpus) bind(c, name="set_script_multigpu")
+    use iso_c_binding, only : c_ptr, c_size_t, c_char, c_int
+    import :: enum_kind
+    integer(kind=enum_kind)                   :: set_script_multigpu_c
+    type(c_ptr), value,            intent(in) :: c_client          !< Initialized SmartRedis client
+    character(kind=c_char),        intent(in) :: key(*)            !< The key to use to place the script
+    integer(kind=c_size_t), value, intent(in) :: key_length        !< The length of the key c-string, excluding null
+    character(kind=c_char),        intent(in) :: script(*)     !< The file storing the script
```
billschereriii, line 102: Doc at the end of this line and the next don't line up properly. (Check `set_script_c` to see if this is a copy/paste of problematic source)

tests/fortran/test_fortran_client.py
```diff
@@ -25,23 +25,25 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import pytest
-from os import path as osp
+import os
```
billschereriii, line 28: Not sure why you've made this change. This driver now diverges from the other driver scripts.  Suggest you either revert it or update the other scripts.

tests/fortran/test_fortran_client.py
```diff
@@ -91,3 +93,17 @@ def execute_cmd(cmd_list):
         print("OUTPUT:", output.decode("utf-8"))
         print("ERROR:", errs.decode("utf-8"))
         assert(False)
+
+@pytest.mark.skipif(
+    not test_gpu,
+    reason="SMARTREDIS_TEST_DEVICE does not specify 'gpu'"
```
billschereriii, line 99: We should document this environment variable somewhere

utils/create_cluster/slurm_cluster.py
```diff
@@ -45,8 +45,10 @@ def launch_db(host, port, dpn):
     set log file
     """
 
+    test_device = environ['SMARTREDIS_TEST_DEVICE']
     redis = environ['REDIS_INSTALL_PATH'] + '/redis-server '
-    redisai = "--loadmodule " +  environ['REDISAI_CPU_INSTALL_PATH'] + "/redisai.so "
+    redisai = "--loadmodule " +  environ[f'REDISAI_{test_device.upper()}_INSTALL_PATH'] + "/redisai.so "
+    print(host)
```
billschereriii, line 51: Is this line debugging code?

billschereriii:
Generally looks good.  A few cosmetic issues and a few questions for you to address

src/c/c_client.cpp
ashao: Good catch.

src/fortran/client.F90
ashao: Good catch. Fortran standard is to say that it will return an error code but not segfault. That being said, you're right that we should wrap this in `if ALLOCATED(c_outputs) deallocate(c_outputs)`

tests/fortran/test_fortran_client.py
ashao: Mainly since now we're using both the `os.environ` and `os.path`. That being said, I agree with you that we should unify the convention.

src/fortran/client.F90
billschereriii: Applies in a couple places and in both this and the non-multi version of the routine

utils/create_cluster/local_cluster.py
```diff
@@ -65,7 +65,7 @@ def create_cluster(n_nodes, port):
     # Start servers
     host = '127.0.0.1'
     redis = os.getenv('REDIS_INSTALL_PATH') + '/redis-server'
-    redisai = os.getenv('REDISAI_CPU_INSTALL_PATH') + '/redisai.so '
+    redisai = os.getenv('REDISAI_GPU_INSTALL_PATH') + '/redisai.so '
```
mellis13, line 68: Do we need some kind of switch here depending on the test device to choose between ``REDISAI_CPU_INSTALL_PATH`` and ``REDISAI_GPU_INSTALL_PATH``?

src/c/c_client.cpp
ashao: Thanks for catching this. Double-checked both C/Fortran and it seems to be just `set_script_from_file_multigpu`

src/fortran/client/script_interfaces.inc
ashao: Fixed.

tests/fortran/test_fortran_client.py
ashao: Reverted to `osp`

utils/create_cluster/slurm_cluster.py
ashao: Removed

utils/create_cluster/local_cluster.py
ashao: Thanks for catching that. I've added a line to check for `SMARTREDIS_TEST_DEVICE`, defaulting to 'cpu' if it's not set.

tests/fortran/test_fortran_client.py
ashao: Turns out that most of the environment variables that control the behavior of testing are not in the documentation. I've added a brief description to the SmartRedis docs, but am deferring updates to the main documentation to another ticket.

billschereriii: Ooh, good catch.  If you'll put in a placeholder for a section on environment variables, we can fill it in more fully when we clean up documentation in the release cycle.

billschereriii (approved):
Looks great, thanks for taking care of this!

PR #256, opened 2022-05-19
Python interface for Dataset Aggregation

billschereriii:
Implement dataset aggregation interface in Python and create unit tests for it

### DIFF
--- a/include/pyclient.h
+++ b/include/pyclient.h
@@ -605,6 +605,29 @@ class PyClient
         void use_model_ensemble_prefix(bool use_prefix);
 
         /*!
+        *   \brief Control whether aggregation lists are prefixed
+        *   \details This function can be used to avoid key collisions in an
+        *            ensemble by prepending the string value from the
+        *            environment variable SSKEYIN and/or SSKEYOUT to
+        *            aggregation list names.  Prefixes will only be used if
+        *            they were previously set through the environment variables
+        *            SSKEYOUT and SSKEYIN. Keys for aggregation lists created
+        *            before this function is called will not be retroactively
+        *            prefixed. By default, the client prefixes aggregation
+        *            list keys with the first prefix specified with the SSKEYIN
+        *            and SSKEYOUT environment variables.  Note that
+        *            use_tensor_ensemble_prefix() controls prefixing
+        *            for the entities in the aggregation list, and
+        *            use_tensor_ensemble_prefix() should be given the
+        *            same value that was used during the initial
+        *            setting of the DataSet into the database.
+        *  \param use_prefix If set to true, all future operations
+        *                    on aggregation lists will use
+        *                    a prefix, if available.
+        */
+        void use_list_ensemble_prefix(bool use_prefix);
+
+        /*!
         * \brief Set whether names of tensors or datasets should be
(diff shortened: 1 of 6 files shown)


include/pyclient.h
```diff
@@ -720,6 +720,172 @@ class PyClient
         */
         void save(std::vector<std::string> addresses);
 
+        /*!
```
mellis13, line 723: Are we missing the ``use_list_ensemble_prefix`` method in this file?

include/pyclient.h
```diff
@@ -720,6 +720,172 @@ class PyClient
...
+        *                    starting at zero).  Negative values are
+        *                    supported.  A negative value indicates offsets
+        *                    starting at the end of the list. For example, -1 is
+        *                    the last element of the list.
+        *   \returns A vector containing DataSet objects.
+        *   \throw SmartRedis::Exception if retrieval fails or
+        *          input parameters are invalid
+        */
+        py::list get_dataset_list_range(const std::string& list_name,
+                                                    const int start_index,
```
mellis13, line 886: Fix indent

src/python/src/pyclient.cpp
```diff
@@ -1113,4 +1114,237 @@ void PyClient::save(std::vector<std::string> addresses)
...
+}
+
+// Get datasets from an aggregation list
+py::list PyClient::get_datasets_from_list(const std::string& list_name)
+{
+    try {
+        std::vector<DataSet> datasets = _client->get_datasets_from_list(list_name);
+        std::vector<PyDataset*> result;
+        for (auto it = datasets.begin(); it != datasets.end(); it++) {
+            DataSet* ds = new DataSet(*it);
```
mellis13, line 1299: Can we use move semantics here to avoid an extra underlying copy of all of the data.  It looks like we are just copying to extend the memory lifetime?n. This comment would extend to the other ``get`` method for aggregation lists.

src/python/src/pyclient.cpp
```diff
@@ -1113,4 +1114,237 @@ void PyClient::save(std::vector<std::string> addresses)
...
+        throw;
+    }
+    catch (std::exception& e) {
+        // should never happen
+        throw SRInternalException(e.what());
+    }
+    catch (...) {
+        // should never happen
+        throw SRInternalException("A non-standard exception was encountered "\
+                                  "while executing copy_list.");
```
mellis13, line 1346: Wrong function name in the error message.

tests/python/test_dataset_aggregation.py
```diff
@@ -0,0 +1,139 @@
+# BSD 2-Clause License
```
mellis13, line 1: Because there is a copy operation (or potentially move if we update the routines) outside of the core C++ layer, should we test in this test that the retrieved datasets are in fact the same as the sent ones?  It might be a good sanity check.

mellis13 (commented):
Thanks for the great PR.  Just a couple of comments/questions before merging.

include/pyclient.h
billschereriii: Good catch!

mellis13 (approved):
Thanks for implementing the Python interface!

PR #257, opened 2022-05-23
C interface for dataset aggregation

billschereriii:
Implement a C interface for dataset aggregation

### DIFF
--- a/include/c_client.h
+++ b/include/c_client.h
@@ -1386,11 +1386,11 @@ SRError rename_list(void* c_client,
 *   \param list_name The list name
 *   \param list_name_length The size in bytes of the list name,
 *                           including null terminator
-*   \param result Receives the length of the list
+*   \param result_length Receives the length of the list
 *   \return Returns SRNoError on success or an error code on failure
 */
 SRError get_list_length(void* c_client, const char* list_name,
-                        const size_t list_name_length, int* result);
+                        const size_t list_name_length, int* result_length);
 
 /*!
 *   \brief Poll list length until length is equal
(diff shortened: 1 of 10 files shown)


tests/python/test_dataset_aggregation.py
```diff
@@ -0,0 +1,157 @@
...
+import numpy as np
+import pytest
+from smartredis import Client, Dataset
+from smartredis.error import *
+
+
+def test_aggregation(use_cluster):
+    client = Client(None, use_cluster)
+
+    # Build datasets
```
ashao, line 37: Could do this in a list e.g.
```
datasets = [create_dataset(f"dataset_{i+1}" for i in range(4)]
```

tests/python/test_dataset_aggregation.py
```diff
@@ -0,0 +1,157 @@
...
+    client.append_to_list(list_name, dataset_1)
+    client.append_to_list(list_name, dataset_2)
+
+    # Put the final two datasets into the list
+    client.put_dataset(dataset_3)
+    client.put_dataset(dataset_4)
+    client.append_to_list(list_name, dataset_3)
+    client.append_to_list(list_name, dataset_4)
+
+    actual_length = 4
```
ashao, line 59: Move this magic number to a default argument to the signature so that we can scale this test if we ever run into the problem for large numbers of datasets

ashao (approved):
Some minor changes, the only substantive thing would be to slightly refactor the python test to allow it to be more scaleable.

include/c_client.h
```diff
@@ -1280,6 +1280,248 @@ SRError use_tensor_ensemble_prefix(void* c_client, bool use_prefix);
...
+*            exist (i.e. this is the first entry in the list).
+*            Aggregation lists work by referencing the dataset
+*            by storing its key, so appending a dataset
+*            to an aggregation list does not create a copy of the
+*            dataset.  Also, for this reason, the dataset
+*            must have been previously placed into the database
+*            with a separate call to put_dataset().
+*   \param c_client The client object to use for communication
+*   \param list_name The name of the aggregation list
+*   \param list_name_length The size in bytes of the list name,
```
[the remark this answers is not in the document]
billschereriii: Done

tests/python/test_dataset_aggregation.py
billschereriii: Done

tests/python/test_dataset_aggregation.py
billschereriii: Can't meaningfully put it in the signature since this is a pytest test case, but I moved it to the top of the function as a single constant that's used throughout the test case so that updates can be made in one place

PR #258, opened 2022-06-03
Implement dataset aggregation for Fortran

ashao:
Adds in all methods and C interfaces to enable the dataset aggregation functionality in Fortran. Most of the transformations and implementations are similar to those done before. The one departure from the C implementation is that for returning a list of datasets, we rely on Fortran to allocate the necessary memory for the list of dataset objects. This vector is then populated within the C client without allocating a new vector.
    
This is just an initial implementation. Tests to follow


src/fortran/client.F90
```diff
@@ -148,8 +149,27 @@ module smartredis_client
 
   procedure :: use_tensor_ensemble_prefix
   procedure :: use_model_ensemble_prefix
+  procedure :: use_list_ensemble_prefix
```
billschereriii, line 152: Missing one-liner descriptions for these

src/fortran/client.F90
```diff
@@ -1444,4 +1464,259 @@ function use_tensor_ensemble_prefix(self, use_prefix) result(code)
...
+
+!> Appends a dataset to the aggregation list When appending a dataset to an aggregation list, the list will
+!! automatically be created if it does not exist (i.e. this is the first entry in the list). Aggregation 
+!! lists work by referencing the dataset by storing its key, so appending a dataset to an aggregation list
+!! does not create a copy of the dataset.  Also, for this reason, the dataset must have been previously 
+!! placed into the database with a separate call to put_dataset().
+function append_to_list(self, list_name, dataset) result(code)
+  class(client_type), intent(in) :: self       !< An initialized SmartRedis client
+  character(len=*),   intent(in) :: list_name    !< Name of the dataset to get
+  type(dataset_type), intent(in) :: dataset !< receives the dataset
```
billschereriii, line 1484: Docstring labels on vars aren't lined up

billschereriii: That appears to apply to many of the new procedures

src/fortran/client.F90
```diff
@@ -1444,4 +1464,259 @@ function use_tensor_ensemble_prefix(self, use_prefix) result(code)
...
+  
+  ! Local variables
+  character(kind=c_char, len=len_trim(list_name)) :: list_name_c
+  integer(kind=c_size_t) :: list_name_length
+  integer(kind=c_int) :: c_poll_frequency, c_num_tries, c_list_length
+
+  list_name_c = trim(list_name)
+  list_name_length = len_trim(list_name)
+  c_num_tries = num_tries
+  c_poll_frequency = poll_frequency_ms
```
billschereriii, line 1585: is this translation necessary? I don't know if we've done it consistently in other places

billschereriii: (from `int` to `c_int`, that is)

src/fortran/client.F90
```diff
@@ -1444,4 +1464,259 @@ function use_tensor_ensemble_prefix(self, use_prefix) result(code)
...
+
+  code = poll_list_length_lte_c(self%client_ptr, list_name_c, list_name_length, &
+                            c_list_length, c_poll_frequency, c_num_tries, poll_result)
+end function poll_list_length_lte
+
+!> Get datasets from an aggregation list. Note that this will deallocate an existing list
+function get_datasets_from_list(self, list_name, datasets, num_datasets) result(code)
+  class(client_type),   intent(in) :: self       !< An initialized SmartRedis client
+  character(len=*),     intent(in) :: list_name    !< Name of the dataset to get
+  type(dataset_type), dimension(:), allocatable, intent(  out) :: datasets !< The array of datasets included
```
billschereriii, line 1646: ( out) -- while I like this convention of putting the space to show the direction, you haven't been doing it consistently

src/fortran/client.F90
```diff
@@ -1444,4 +1464,259 @@ function use_tensor_ensemble_prefix(self, use_prefix) result(code)
...
+  code = self%get_list_length(list_name, num_datasets)
+  allocate(dataset_ptrs(num_datasets))
+
+  list_name_c = trim(list_name)
+  list_name_length = len_trim(list_name)
+  
+  c_num_datasets = num_datasets
+  code = get_dataset_list_range_allocated_c(self%client_ptr, list_name_c, list_name_length, &
+                                           0, c_num_datasets-1, dataset_ptrs) 
+
```
billschereriii, line 1667: Splitting the call like this requires an extra round trip to the database. I don't think that this routine is going to be called enough that performance is critical, but that is a consideration

src/fortran/client/aggregation_interfaces.inc
```diff
@@ -0,0 +1,173 @@
...
+  end function get_datasets_from_list_c
+end interface
+
+interface
+  function get_dataset_list_range_allocated_c(client, list_name, list_name_length, start_index, end_index, &
+    datasets) bind(c, name="_get_dataset_list_range_allocated")
+    use iso_c_binding, only : c_ptr, c_char, c_bool, c_size_t, c_int
+    import :: enum_kind
+    integer(kind=enum_kind)       :: get_dataset_list_range_c
+    type(c_ptr), value            :: client
```
billschereriii, line 166: This file looks fine except for the way the docstrings line up

include/c_client.h
```diff
@@ -1280,6 +1280,283 @@ SRError use_tensor_ensemble_prefix(void* c_client, bool use_prefix);
...
+*   \param end_index The ending index of the range (inclusive,
+*                    starting at zero).  Negative values are
+*                    supported.  A negative value indicates offsets
+*                    starting at the end of the list. For example, -1 is
+*                    the last element of the list.
+*   \param datasets Receives an array of datasets included in the list
+*   \param num_datasets Receives the number of datasets returned
+*   \return Returns SRNoError on success or an error code on failure
+*/
+SRError _get_dataset_list_range_allocated(void* c_client, const char* list_name,
```
billschereriii, line 1555: This needs to say somewhere that it is ONLY for use by the Fortran interface

src/c/c_client.cpp
```diff
@@ -1674,3 +1674,389 @@ SRError use_tensor_ensemble_prefix(void* c_client, bool use_prefix)
...
+    if ( num_datasets != (end_index-start_index+1)) {
+      SRSetLastError(SRInternalException(
+        "Returned dataset list is not equal to the requested range"
+      ));
+    }
+
+    if (num_datasets > 0) {
+      for (size_t i = 0; i < num_datasets; i++) {
+        datasets[i] = (void*)(new DataSet(std::move(result_datasets[i])));
+      }
```
billschereriii, line 2049: If this routine doesn't work, change the declaration to a void*** and make this line:
`(*datasets)[i] = (void*)(new DataSet(std::move(result_datasets[i])));`

billschereriii (approved):
Generally looks good. Approved, conditioned on completing the cosmetic changes documented and on getting a clean pass through the check-in tests.

src/fortran/client.F90
ashao: Fixed

src/fortran/client.F90
ashao: We talked about this offline, but for posterity: for the sake of absolute best practices, we should be much better about explicitly converting from the default integer type that the compiler uses to a specific c-integer. There are a number of cases where the users might specify that all integers/reals/etc. should 64-bit numbers where we might run into trouble if we pass them straight to the c client.

src/fortran/client.F90
ashao: That's a good point. I'll add a comment in the code here to make sure we don't forget about this if it comes up.

include/c_client.h
ashao: Good point. I don't see any reason why it couldn't be used on the C side, but I agree that we should keep it semi-private with disclaimers for now.

src/c/c_client.cpp
ashao: This worked fine for the Fortran implementation which is somewhat baffling to me. I think perhaps because the memory is allocated on the Fortran side, and then we're passing a `c_ptr` (essentially a `void *`) to the Fortran array. Since Fortran is passing everything by reference, then it is correctly pointing to the first element of the Fortran array after which C is just striding through memory at the size of a `void *`.

PR #259, opened 2022-06-09
Add documentation of new environment variables

billschereriii:
Update documentation to detail new environment variables.

### DIFF
--- a/doc/runtime.rst
+++ b/doc/runtime.rst
@@ -110,11 +110,10 @@ The functions for changing this default behavior are:
 Model Execution Environment Variable
 ====================================
 
-SmartRedis enables a timeout on the length of time that is allowed
-to elapse while waiting for a model to elapse. This is controlled
-by the ``SR_MODEL_TIMEOUT`` environment variable. The value for
-this variable is measured in milliseconds, and the default value
-is one minute.
+The ``SR_MODEL_TIMEOUT`` environment variable defines a timeout
+on the length of time SmartRedis will wait for a model to
+execute. The value for this variable is measured in milliseconds,
+and the default value is one minute.
 
 Connection and Command Execution Environment Variables
 ======================================================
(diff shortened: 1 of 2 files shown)


doc/runtime.rst
```diff
@@ -107,6 +107,15 @@ The functions for changing this default behavior are:
     The function ``Client.use_model_ensemble_prefix()`` controls
     object prefixing for model and script data.
 
+Model Execution Environment Variable
+====================================
+
+SmartRedis enables a timeout on the length of time that is allowed
+to elapse while waiting for a model to elapse. This is controlled
```
mellis13, line 114: should this be "waiting for a model to execute."?.

mellis13 (approved):
Thanks for the update!  Feel free to merge after group discussion on defaults.

doc/runtime.rst
```diff
@@ -107,6 +107,15 @@ The functions for changing this default behavior are:
     The function ``Client.use_model_ensemble_prefix()`` controls
     object prefixing for model and script data.
 
+Model Execution Environment Variable
+====================================
+
+SmartRedis enables a timeout on the length of time that is allowed
```
ashao, line 113: I might suggest leading with the variable name, i.e. `SR_MODEL_TIMEOUT` sets the amount of time that the client will wait for a model to execute [...]

doc/runtime.rst
```diff
@@ -130,3 +139,11 @@ The user can set these environment variables to adjust command execution behavio
 ``SR_CMD_TIMEOUT`` should be specified in seconds.  Note that ``SR_CMD_INTERVAL``
 and ``SR_CMD_TIMEOUT`` are read during client initialization and not
```
ashao, line 140: Isn't this behavior true of all of the environment variables for the client?

ashao (changes requested):
Very minor comments for clarity. Thank you @billschereriii for making sure these were documented.

doc/runtime.rst
billschereriii: We don't currently have an overview of environment variables, but maybe we should add a table to the top of this text that has overview info like this. Not something that can fit into this cycle though

ashao (approved):

PR #260, opened 2022-06-13
Fortran docs

billschereriii:
De-macro procedure definitions to make source code more compatible with Doxygen

### DIFF
--- a/src/fortran/client.F90
+++ b/src/fortran/client.F90
@@ -408,6 +408,11 @@ end function poll_key
 !> Put a tensor whose Fortran type is the equivalent 'int8' C-type
 function put_tensor_i8(self, name, data, dims) result(code)
   integer(kind=c_int8_t), dimension(..), target, intent(in) :: data !< Data to be sent
+  class(client_type),                    intent(in) :: self !< Fortran SLIC client
+  character(len=*),                      intent(in) :: name !< The unique name used to store in the database
+  integer, dimension(:),                 intent(in) :: dims !< The length of each dimension
+  integer(kind=enum_kind)                           :: code
+
   include 'client/put_tensor_methods_common.inc'
 
   ! Define the type and call the C-interface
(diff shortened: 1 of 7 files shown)


src/fortran/client.F90
```diff
@@ -408,6 +408,11 @@ end function poll_key
 !> Put a tensor whose Fortran type is the equivalent 'int8' C-type
 function put_tensor_i8(self, name, data, dims) result(code)
   integer(kind=c_int8_t), dimension(..), target, intent(in) :: data !< Data to be sent
+  class(client_type),                    intent(in) :: self !< Fortran SLIC client
```
ashao, line 411: SLIC/SILC is a holdover from the early days of SmarTsim. Maybe rename to "An initialized SmartRedis Fortran client"

tests/cpp/test_cpp_client.py
```diff
@@ -62,7 +62,7 @@ def execute_cmd(cmd_list):
     proc = Popen(
         cmd_list, stderr=PIPE, stdout=PIPE, stdin=PIPE, cwd=run_path)
     try:
-        out, err = proc.communicate(timeout=120)
+        out, err = proc.communicate(timeout=180)
```
ashao, line 65: Any reason in particular why this was increased?

ashao (changes requested):
Other than a minor change in the description and a question about the timeout in the CPP tests, this is pretty much good to go. Thanks @billschereriii

tests/cpp/test_cpp_client.py
ashao: Also, we should make sure we make these magic numbers more obvious.

src/fortran/client.F90
billschereriii: Sure. Didn't even look at it when I copy/pasted it

tests/cpp/test_cpp_client.py
billschereriii: The CPP unit tests are timing out occasionally in the checkin, so I bumped it to 3 minutes from 2, Should really have done this in a separate PR but I was being lazy.  I agree in general about magic numbers, but this is the driver for our test cases, so not production code

ashao: Agreed. Mainly just a note to myself to think about how to consolidate these kinds of things.

ashao (approved):
Thanks @billschereriii. Feel free to merge

PR #270, opened 2022-07-06
Add timing info for the beginning of each test case in CPP unit tests

billschereriii:
Enables identification of which tests are taking a really long time in check-in testing


billschereriii:
Not seeing any weird timing data on these runs, so I suggest we merge this in and hopefully we'll be able to find out what's going on in some future run

ashao (approved):
We might want to revisit setting these as per-test timers versus an accumulated timer, but this will be great to solve some of our more immediate problems with timeouts.

PR #277, opened 2022-08-23
Update linking instructions and test builds

ashao:
https://github.com/CrayLabs/SmartRedis/pull/245 introduced the ability to compile the Fortran
clients into their own shared libraries that Fortran applications can build against. Here we
include documentation to guide users to incorporate the new library into their own build
systems. This includes specific instructions for the CMake-based build systems as well
as compiler flags at both the preprocessing and link stages. Lastly, the Fortran tests
now link against the Fortran library, slightly decreasing the build time.


.github/workflows/run_tests_uds.yml
```diff
@@ -142,5 +142,7 @@ jobs:
 
       # CI tests that should be run with UDS Redis go here
       - name: Run testing with UDS redis
-        run: make test-verbose
+        run: |
+          export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/install/lib &&
+          make test-verbose
```
billschereriii, line 147: I suspect you need to make this change to the post-merge tests yml file as well

doc/changelog.rst
```diff
@@ -14,20 +14,22 @@ Description
...
 - This version adds new functionality in the form of support for Unix Domain Sockets.
 - Fortran client can now be optionally built with the rest of the library
-- Initial support for dataset conversions, specifically Xarray. 
+- Initial support for dataset conversions, specifically Xarray.
 
 Detailed Notes
 
 - Implemented support for Unix Domain Sockets, including refactorization of server address code, test cases, and check-in tests. (PR252_)
-- A new make target `make lib-with-fortran` now compiles the Fortran client and dataset into its own library which applications can link against (245_)
+- A new make target `make lib-with-fortran` now compiles the Fortran client and dataset into its own library which applications can link against (PR245_)
```
billschereriii, line 22: These should be in reverse chronological order to be consistent with earlier changelogs

src/fortran/client.F90
```diff
@@ -200,28 +203,32 @@ function SR_error_parser(self, response_code) result(is_error)
...
+        write(stderr,*) "Timeout error"
+      case(SRKeyError)
+        write(stderr,*) "Key error"
+      case(SRTypeError)
+        write(stderr,*) "Type mismatch error"
+      case default
+        write(stderr,*) "Invalid or uninitialized response code"
+    end select
+  else
+    write(stderr,*) "SmartRedis Client has not been initialized"
```
billschereriii, line 230: Good catch!

billschereriii (approved):
Some good updates to clean up the Fortran and bring it to standard. A few minor comments to address, but nothing worth holding up approval over

mellis13 (changes requested):
Thanks for this PR -- it will be really nice to have fortran users be able to link against a fortran library.

doc/install/from_source.rst
```diff
@@ -1,191 +0,0 @@
-
-Developer Makefile
```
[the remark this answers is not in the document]
ashao: Yes, it's because of redundant information. I'll submit a SmartSim PR to update.

mellis13 (approved):
Looks good to merge.  Let's just make an issue to update SmartSim docs because of the deleted file and address that next.

PR #280, opened 2022-11-16
Initialize boolean before starting thread pool

mellis13:
This PR is a small fix that moves the thread pool ``shutting_down`` boolean to before the initialization of the thread pool.  Shutting down of threads was observed during thread pool creation because the variable was left uninitialized.


billschereriii (approved):
Good catch, should have had it this way from day 1

PR #281, opened 2022-11-16
Logging for SmartRedis

billschereriii:


billschereriii:
Initial C++ implementation is code complete, but not tested (or even compiled!). Provided for team review of implementation direction only

include/utility.h
```diff
@@ -0,0 +1,62 @@
...
+#ifndef SMARTREDIS_UTILITY_H
+#define SMARTREDIS_UTILITY_H
+
+#include <cstdlib>
+#include <string>
+#include "address.h"
+
+///@file
+
+namespace SmartRedis {
```
ankona, line 38: Do we want the env utility code in the root namespace instead of something like `SmartRedis::Util`?

src/cpp/client.cpp
```diff
@@ -63,11 +73,17 @@ Client::~Client()
         _redis = NULL;
     }
     _redis_server = NULL;
+
+    Logger& logger = Logger::get_instance();
```
ankona, line 77: I feel like this style implementation just screams "code churn" even though I know it's effective.

Do you prefer having to add these logEnter/logExit blocks all over or would a decorator object like the following be cleaner? Please forgive my lack of c language skill

```

class LoggedClient : public IClient
   public LoggedClient(Client rootClient) { ... }

   void SomeClientMethod(...) {
      Logger& logger = Logger::get_instance();
      logger.log_data(llInfo, "some enter message");

      _rootClient.SomeClientMethod(...);

      logger.log_data(llInfo, "some exit message");
   }
}
```

This could have the added benefit of injecting the `LoggedClient` or `Client` instance and avoiding having to check every call for your log level or something like `if (loggingEnabled) { doLogging() }`

include/srexception.h
```diff
@@ -101,7 +102,8 @@ class Exception: public std::exception
     Exception(const std::string& what_arg, const char* file, int line)
       : _msg(what_arg), _loc(file + std::string(":") + std::to_string(line))
     {
-        // NOP
+        Logger::get_instance().log_data(
+            LLInfo, exception_class() + " at " + _loc + ":" + _msg);
```
ankona, line 106: Should we add a `log_exception` function to the logger to avoid exception classes having to re-implement this "{typename} at {loc}:{message}" format?

`log_exception` would add the benefit that if you choose to modify your logging format, all of the exception classes wouldn't have to update their implementation.

include/srexception.h
```diff
@@ -223,6 +254,14 @@ class DatabaseException: public Exception
     SRError to_error_code() const noexcept override {
         return SRDatabaseError;
     }
+
+    /*!
+    *   \brief Get a string representation of the exception class
+    *   \returns Stringified version of the class name
+    */
+    virtual std::string exception_class() {
+        return std::string("DatabaseException")
```
ankona, line 263: I had to google this, but would something like [this](https://stackoverflow.com/questions/3649278/how-can-i-get-the-class-name-from-a-c-object) avoid refactor/rename/copypasta errors from reporting incorrect class names?

```
return typeid(theObject).name();
```

src/cpp/client.cpp
```diff
@@ -1366,16 +1552,18 @@ std::vector<DataSet> Client::get_dataset_list_range(const std::string& list_name
 void Client::_set_prefixes_from_env()
 {
     // Establish set prefix
-    const char* keyout_p = std::getenv("SSKEYOUT");
-    if (keyout_p != NULL)
-        _put_key_prefix = keyout_p;
+    std::string put_key_prefix;
+    get_string_from_env(put_key_prefix, "SSKEYOUT", "");
+    if (put_key_prefix.length() > 0)
+        _put_key_prefix = put_key_prefix;
```
ankona, line 1558: Is there a c++ equivalent of `IConfiguration` that could be passed to the constructor? 

By passing configuration in, you can decouple where the config comes from (and build hierarchical config, use a config file, change env var names without churn, etc).

src/cpp/client.cpp
billschereriii: If you look at the way we've added API tracking, it's only a single line of code added to each API point. What you're describing above duplicates the API points and involves a ton more code being written

include/srexception.h
billschereriii: We only have to do this in one place. All the other SmartRedis exception classes are getting their implementation through polymorphic inheritence

src/cpp/client.cpp
billschereriii: Let's chat -- I'm not sure I understand what you're suggesting here

include/srexception.h
```diff
@@ -191,6 +201,19 @@ class Exception: public std::exception
...
+    *   \brief BadAllocException constructor with location information
+    *   \param what_arg The message for the exception
+    *   \param file The source file from which the exception was thrown
+    *   \param line The line number from which the exception was thrown
+    */
+    BadAllocException(const std::string& what_arg, const char* file, int line)
+      : Exception(what_arg, file, line)
+    {
+        Logger::get_instance().log_data(
+            LLInfo, "BadAllocException at " + _loc + ":" + _msg);
```
ankona, line 214: Per polymorphic comment above - is this going to double up on the formatting since BadAllocEx sends a "location-ified string" to the Exception constructor that already logs `Logger::get_instance().log_data(...)`?

include/srexception.h
billschereriii: GCC gives back a mangled name for this, which will not be very user friendly, so this isn't going to be viable.

include/sr_enums.h
```diff
@@ -72,4 +72,14 @@ typedef enum {
     SRTensorTypeUint16  = 8  // 16-bit unsigned integer tensor type
 } SRTensorType;
 
+/*!
+*   \brief  Enumeration for logging levels
+*/
+typedef enum {
+    LLInvalid = 0, // Invalid or uninitialized logging level
```
mellis13, line 79: SmartSim progresses with:

Quiet
Info
Debug
Developer

It might be nice to be consistent with SmartSim so users and developers don't have to remember two different lexicons.

include/logger.h
```diff
@@ -0,0 +1,175 @@
...
+        *   \param data Text of data to be logged
+        */
+        void log_data(SRLoggingLevel level, const std::string_view& data);
+
+    private:
+
+        /*!
+        *   \brief Track whether logging is initialized
+        */
+        bool initialized;
```
mellis13, line 116: In other files I think we have been using ``_`` before private member variables.  It might be nice to stay consistent with other conventions.  This is a general comment and won't comment all new classes to avoid redundant comments.

mellis13 (changes requested):
Thanks for this draft PR!  I did a high level review and just had a couple of comments/questions.  I can do a more detailed review when it is done.

include/logger.h
billschereriii: Done, thanks for noticing that I got the convention backwards here

include/sr_enums.h
billschereriii: Agree 100% that it makes sense to stay consistent. I've adjusted the levels to match SmartSim

ashao (changes requested):
Really great implementation and you did a great job keeping it very compact. I really appreciate the refactor to isolate the places where we get the values of configuration variables.

The largest comment/criticism that I have is that the logging is quite closely tied to the `Client`. As we've discussed offline, this might interfere with multiple client support down the line, but after reviewing this PR there's a larger question about how the logging interacts with various entities within SmartRedis. My suggestion is that we pass in the SmartRedis object that we're trying to log and that the identifier that appears in the logging file is extracted from the object. For example:
```
log_message(dataset, "Corralling llamas")
log_message(client, "Reticulating splines")
```
with the resulting logfile saying
dataset-1: Corralling llamas
client-foo: Reticulating splines

In this way, the Client, Dataset, and Logger are all first-class entities (which interact with each other) whereas now the Logger is somewhat subordinate to Client. This can be seen most clearly in the Dataset constructor which allows you to set a client id. Happy to talk about this in greater detail

examples/serial/python/example_model_file_torch.py
```diff
@@ -43,7 +43,7 @@ def forward(self, x):
 
 # Connect a SmartRedis client
 db_address = "127.0.0.1:6379"
-client = Client(address=db_address, cluster=True)
+client = Client(address=db_address, cluster=True, client_id="example_model_file_torch.py")
```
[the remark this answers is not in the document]
billschereriii: __file__ gives the full path, which is annoying to see in the logs. I actually moved away from using it in the Fortran and C++ examples and test cases for this reason

include/logger.h
```diff
@@ -0,0 +1,288 @@
...
+        /*!
+        *   \brief The name of the current function
+        */
+        std::string name;
+};
+
+/*!
+*   \brief Instantiate a FunctionLogger for the enclosing API function
+*/
+#define LOG_API_FUNCTION() FunctionLogger ___function_logger___(__func__)
```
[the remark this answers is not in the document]
billschereriii: Tweaking the behavior of the FunctionLogger is certainly easy enough

doc/changelog.rst
```diff
@@ -13,21 +13,26 @@ This section details changes made in the development branch that have not yet be
 Description
 
 - Python error messages from SmartRedis contain more information
+- Added logging functionality to the SmartRedis library
 - A bug related to thread pool initialization was fixed.
 - This version adds new functionality in the form of support for Unix Domain Sockets.
 - Fortran client can now be optionally built with the rest of the library
 - Initial support for dataset conversions, specifically Xarray.
+- Added support for logging of client activity.
```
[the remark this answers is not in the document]
billschereriii: Fixed, good catch

src/fortran/logger/logger_interfaces.inc
```diff
@@ -0,0 +1,55 @@
...
+    integer(kind=c_size_t), value  :: data_length
+  end subroutine c_log_data
+end interface
+
+interface
+  subroutine c_log_warning(level, data, data_length) bind(c, name="log_warning_noexcept")
+    use iso_c_binding, only : c_ptr, c_char, c_size_t
+    import :: enum_kind
+    integer(kind=enum_kind), value :: level
+    character(kind=c_char)         :: data(*)
```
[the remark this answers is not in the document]
billschereriii: Doesn't seem to work

src/fortran/logger/logger_interfaces.inc
```diff
@@ -0,0 +1,55 @@
...
+    integer(kind=c_size_t), value  :: data_length
+  end subroutine c_log_warning
+end interface
+
+interface
+  subroutine c_log_error(level, data, data_length) bind(c, name="log_error_noexcept")
+    use iso_c_binding, only : c_ptr, c_char, c_size_t
+    import :: enum_kind
+    integer(kind=enum_kind), value :: level
+    character(kind=c_char)         :: data(*)
```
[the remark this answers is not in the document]
billschereriii: Doesn't seem to work

src/cpp/client.cpp
```diff
@@ -29,20 +29,30 @@
...
 
 // Constructor
-Client::Client(bool cluster)
-    : _redis_cluster(cluster ? new RedisCluster() : NULL),
-      _redis(cluster ? NULL : new Redis())
+Client::Client(bool cluster, const std::string& client_id)
 {
+    // Set up logging
+    Logger& logger = Logger::get_instance();
+    logger.configure_logging(client_id);
```
mellis13, line 42: Is there a way that we can make logger id not dependent on the last constructor called?  If we wanted to log from other objects that were just built, this seems like it would be confusing/difficult to maintain and read.

mellis13 (changes requested):
Great PR! This is a huge step in the right direction for the code.  I had similar general questions as Andrew regarding code structure, but won't repeat here.  I'd be happy to join follow-up design discussions.

Just as a final note, we should run the scaling tests before merging to make sure there is no perf hit.

include/logger.h
```diff
@@ -0,0 +1,288 @@
...
+#endif // __cplusplus
+#include <stdlib.h>
+#include "sr_enums.h"
+
+///@file
+
+/*
+* Redirect logging functions to exception-free variants for C and Fortran
+*/
+#ifndef __cplusplus
```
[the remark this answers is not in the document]
billschereriii: Honestly, I don't think this would really affect the readability. It would also be bad to have the implementations split up between a c_logger.cpp and logger.cpp -- there just isn't enough content to merit splitting up the file

include/logger.h
```diff
@@ -0,0 +1,288 @@
...
+        *   \param _client_id new ID to use for the current client
+        */
+        void rename_client(const std::string& client_id);
+
+    private:
+
+        /*!
+        *   \brief Logger constructor
+        */
+        Logger() { _initialized = false; }
```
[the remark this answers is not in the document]
billschereriii: The usual rule of thumb is that inline header definitions are ok if it's one or two lines of code. rename_client could be moved into the header, but it's not necessary to inline things

include/logger.h
```diff
@@ -0,0 +1,288 @@
...
+        */
+       static Logger& get_instance()
+       {
+           static Logger __instance; // instantiated on first acccess
+           return __instance;
+       }
+
+        /*!
+        *   \brief Set up logging for the current client
+        *   \param _client_id ID to use for the current client
```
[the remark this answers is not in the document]
billschereriii: good catch, thanks. I'll do this before the global change to the parameter name

include/logger.h
```diff
@@ -0,0 +1,288 @@
...
+        */
+       void configure_logging(const std::string& client_id);
+
+        /*!
+        *   \brief Rename the current client
+        *   \param _client_id new ID to use for the current client
```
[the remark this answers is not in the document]
billschereriii: good catch, thanks. I'll do this before the global change to the parameter name

include/logger.h
```diff
@@ -0,0 +1,288 @@
...
+        *   \param logger The Logger to copy for construction
+        */
+        void operator=(const Logger& logger) = delete;
+
+        /*!
+        *   \brief Default Logger destructor
+        */
+        ~Logger() = default;
+
+    public:
```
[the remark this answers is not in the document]
billschereriii: Generally, it's accepted practice to put constructors, destructors, and operator overloads at the top of a header; and to put member variables at the bottom.  This follows that practice and the scoping labels follow as needed

include/logger.h
```diff
@@ -0,0 +1,288 @@
...
+        */
+        ~FunctionLogger()
+        {
+            log_data(LLDebug, "API Function " + name + " exited");
+        }
+    private:
+        /*!
+        *   \brief The name of the current function
+        */
+        std::string name;
```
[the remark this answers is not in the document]
billschereriii: Good catch, will fix

doc/runtime.rst
```diff
@@ -42,6 +42,31 @@ location.  However, the Python ``Client`` constructor also allows
...
 this case, it sets SSDB from the input parameter.
 
+Logging Environment Variables
+=============================
+
+SmartRedis will log events to a file on behalf of a client. There
+are two main environment variables that affect logging, ``SR_LOG_FILE``
+and ``SR_LOG_LEVEL``.
+
+``SR_LOG_FILE`` is the specifier for the location of the file that
```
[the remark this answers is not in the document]
billschereriii: Good call, will update the docs to discuss

doc/runtime.rst
```diff
@@ -42,6 +42,31 @@ location.  However, the Python ``Client`` constructor also allows
...
+receives logging information. Each entry in the file will be prefixed
+with a timestamp and the identifier of the client that invoked the logging
+message.
+
+``SR_LOG_LEVEL`` relates to the verbosity of information that wil be logged.
```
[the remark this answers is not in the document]
billschereriii: Yes, will update the docs to discuss

doc/runtime.rst
```diff
@@ -42,6 +42,31 @@ location.  However, the Python ``Client`` constructor also allows
...
+transpire within the SmartRedis library and creation or destruction of a
+client object.  ``DEBUG`` provides more verbose logging, including information
+on the activities of the SmartRedis thread pool and API function entry and exit.
+Debug level logging will also log the absence of an expected environment variable,
+though this can happen only if the variables to set up logging are in place.
+
+The runtime impact of log levels NONE or INFO should be minimal on
+client performance; however, seting the log level to DEBUG may cause some
+degradation.
+
```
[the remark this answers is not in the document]
billschereriii: Good call, will update the docs to discuss

src/cpp/logger.cpp
```diff
@@ -0,0 +1,252 @@
...
+            _logfile = "";
+    }
+
+    // Get the logging level
+    std::string level;
+    get_config_string(level, "SR_LOG_LEVEL", "", true);
+    bool missingLogLevel = level.length() == 0;
+    bool badLogLevel = false;
+    if (level.length() > 0) {
+        if (level.compare("QUIET") == 0)
```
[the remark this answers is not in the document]
billschereriii: I kept going back and forth on this. It's probably more customer friendly to do a to_lower()?

src/cpp/logger.cpp
```diff
@@ -0,0 +1,252 @@
...
+            _log_level = LLInfo;
+        }
+    }
+    else {
+        _log_level = LLInfo;
+    }
+
+    // Now that everything is configured, issue warning and
+    // error messages
+    if (missingLogFile) {
```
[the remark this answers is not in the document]
billschereriii: Yes! It gives us a chance to fully establish logging and log some errors the way the customer wants them logged instead of in our default method. This in turn will improve the chances of the errors being noticed.

src/cpp/logger.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (level > _log_level)
+        return;
+
+    // Get the current timestamp
+    auto t = std::time(NULL);
+    auto tm = *std::localtime(&t);
+    auto timestamp = std::put_time(&tm, "%H-%M-%S");
+
+    // write the log data
+    // (There must be a cleaner way to write this!)
```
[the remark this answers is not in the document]
billschereriii: I tried the ostream approach -- that didn't work (for reasons I don't recall now).
I like the idea of making this into a ticket and will do that

src/cpp/logger.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (_logfile.length() > 0) {
+        std::ofstream logstream;
+        logstream.open(_logfile, std::ios_base::app);
```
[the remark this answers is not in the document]
billschereriii: We can't keep the file open because another writer might need to open the file to write it and I think that triggers a filesystem lock.

You're right that I should be checking good() here and I will add that

tests/cpp/unit-tests/CMakeLists.txt
```diff
@@ -98,6 +99,8 @@ set(UNIT_TESTS
 	test_dbinfocommand.cpp
 	test_clusterinfocommand.cpp
     test_redisserver.cpp
```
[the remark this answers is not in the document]
billschereriii: Will fix

tests/cpp/unit-tests/test_logger.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "logger.h"
+
+unsigned long get_time_offset();
+
+using namespace SmartRedis;
+
+SCENARIO("Additional Testing for logging", "[LOG]")
+{
+    std::cout << std::to_string(get_time_offset()) << ": Additional Testing for logging" << std::endl;
+    Logger::get_instance().rename_client("test_logger");
```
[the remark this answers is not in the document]
billschereriii: I considered it, but it's actually very tricky to get right and I wanted to get the PR submitted since it had gone on so long. We can add more tests as a follow-on PR

src/cpp/client.cpp
billschereriii: That would be a major change. I see what you're saying though. Let's discuss this in conjunction with some of Andrew's feedback

include/client.h
```diff
@@ -73,10 +74,11 @@ class Client
         /*!
         *   \brief Client constructor
         *   \param cluster Flag for if a database cluster is being used
+        *   \param client_id Name to use for this client when logging
         *   \throw SmartRedis::Exception if client connection or
         *          object initialization fails
         */
-        Client(bool cluster);
+        Client(bool cluster, const std::string& client_id = "anonymous");
```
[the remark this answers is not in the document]
ankona: would hostname (something that does offer some value instead of a constant) make sense?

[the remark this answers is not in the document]
billschereriii: Switched to "default"

src/cpp/logger.cpp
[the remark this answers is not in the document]
billschereriii: I found another approach that works reasonably cleanly

src/cpp/logger.cpp
```diff
@@ -0,0 +1,252 @@
...
+        logstream << _client_id << "@" << timestamp << ":" << data
+                  << std::endl;
+    } else {
```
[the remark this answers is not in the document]
billschereriii: No longer relevant with rewrite

src/cpp/logger.cpp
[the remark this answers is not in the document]
billschereriii: Added check for good(). On failure, roll over logging to console

tests/python/test_logging.py
```diff
@@ -0,0 +1,49 @@
...
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import os
+
+import numpy as np
+import pytest
+from smartredis import *
```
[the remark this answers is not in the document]
billschereriii: More seriously, we need Client, log_data, log_warning, log_error, and all four logging level enum constants. That's way too much to list out one at a time

src/cpp/client.cpp
billschereriii: Marking this as resolved since it's largely the same as Andrew's comment and we have an architectural direction

src/cpp/dataset.cpp
```diff
@@ -29,6 +29,7 @@
 #include <string_view>
 #include "dataset.h"
 #include "srexception.h"
+#include "logger.h"
```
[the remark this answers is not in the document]
billschereriii: Per group discussion today, the following rearchitectural changes will resolve this:
 - Creatie SRObject base class for Client and DataSet
 - SRObject contains logging name and log methods
 - Add LogContext class derived from SRObject to allow creation of user-defined contexts for logging in support of user-level logging
 - Log methods in logger.cpp receive an SRObject as the log context; the name is used as the prefix to the log entries
 - Accept NULL as the context, using generic "SmartRedis library" as the log entry prefix for exceptions and other internal places where fuller context is unavailable
 - Pass Client pointer to Redis server and ThreadPool classes to use as context for them

tests/python/test_errors.py
```diff
@@ -740,6 +740,27 @@ def test_bad_type_get_dataset_list_range(use_cluster):
...
+        log_data(LLInfo, 42)
+
+def test_bad_type_log_warning(use_cluster, context):
+    c = Client(None, use_cluster, logger_name=context)
+    with pytest.raises(TypeError):
+        log_warning("Not a logging level", "Data to be logged")
+    with pytest.raises(TypeError):
+        log_warning(LLInfo, 42)
+
+def test_bad_type_log_error(use_cluster, context):
```
MattToast, line 757: at risk of over complicating the test suite, you could do something like
```py
@pytest.mark.parameterize("log_fn", [
    (log_data,), (log_warning,), (log_error,)
])
def test_bad_log_level(use_cluster, context, log_fn):
    ...
```
so you that the same test doesn't have to get rewritten for any/every new log fns that may get made in future (assuming the all take a log level as the first param)

tests/python/test_logging.py
```diff
@@ -0,0 +1,45 @@
...
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from smartredis import *
+from smartredis.error import *
+
+
+def test_logging(use_cluster, context):
```
MattToast, line 31: incredibly at risk of over complicating the test suite, but very similar to my previous note, you could do

```py
@pytest.mark.parameterize("log_fn, log_level", [
    itertools.product(
        [log_data, log_warning, log_error],
        [LLQuiet, LLInfo, LLDebug, LLDeveloper])
])
def test_logging(use_cluster, context, log_fn, log_level):
    ...
```
So that if one of the cases below were to fail, the rest would still run

MattToast (commented):
PR looks amazing!! I took a look mostly at the python side of things to make sure my feedback was documented were it was most useful. I have a couple of things I wanted to confirm but otherwise looks fantastic!

I'll circle back around tomorrow to look over the rest of it in more detail tomorrow, but scanning over Andrew and Matt's feedback it looks like they have already commented on, and  you have addressed, most of my concerns

tests/python/test_errors.py
billschereriii: good thought -- done

src/python/module/smartredis/__init__.py
```diff
@@ -24,6 +24,25 @@
...
+    "LLDeveloper",
+]
+
 from .client import Client
 from .dataset import Dataset
-from .dataset_utils import DatasetConverter
\ No newline at end of file
+from .dataset_utils import DatasetConverter
+from .logger import (
+    log_data, log_warning, log_error
```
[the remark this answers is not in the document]
billschereriii: Absolutely! user-induced logging is a key part of the feature

src/python/module/smartredis/__init__.py
```diff
@@ -24,6 +24,25 @@
...
+)
+from .smartredisPy import (
+    LLQuiet, LLInfo, LLDebug, LLDeveloper
```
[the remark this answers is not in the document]
billschereriii: Prefer to keep the constants named the same in all clients to avoid confusion

src/cpp/logger.cpp
```diff
@@ -0,0 +1,243 @@
...
+        throw SRRuntimeException(
+            "Cannot write to file: " + requestedLogfile);
+    }
+    if (badLogLevel) {
+        throw SRRuntimeException(
+            "Unrecognized logging level: " + level);
+    }
+
+    // Done
+    _initialized = true;
```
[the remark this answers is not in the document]
billschereriii: Good catch, second initialization deleted

conftest.py
```diff
@@ -32,6 +32,7 @@
 import os
 import random
 import string
+import inspect
```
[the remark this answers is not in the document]
billschereriii: done

tests/python/test_logging.py
billschereriii: I was able to use this idea partially to streamline the test by parameterizing over the log levels. The differences in the log messages make parameterizing over log_fn add more code than it saves

src/cpp/threadpool.cpp
```diff
@@ -58,49 +51,41 @@ void ThreadPool::shutdown()
...
-    #endif
     for (std::thread& thr : threads) {
         cv.notify_all(); // Wake up all the threads
-        #ifdef TIME_THREADPOOL
-        std::cout << "Waiting for a thread to terminate (" << std::to_string(i++) << " of "
-                  << std::to_string(num_threads) << ")" << std::endl;
-        #endif
+        std::string message =
+            "Waiting for thread to terminate (" +
+            std::to_string(i++) + " of " +
```
MattToast, line 66: Can this be changed to `++i` so that it counts 1-to-n as opposed to 0-to-(n-1)?

src/cpp/threadpool.cpp
```diff
@@ -31,9 +26,7 @@ ThreadPool::ThreadPool(unsigned int num_threads)
     // Create worker threads
 	if (num_threads < 1) num_threads = 1; // Force a minimum of 1 thread
     for (unsigned int i = 0; i < num_threads; i++) {
-        #ifdef TIME_THREADPOOL
-        std::cout << "Kicking off thread " + std::to_string(i) << std::endl;
-        #endif
+        log_data(LLDeveloper, "Kicking off thread " + std::to_string(i));
```
MattToast, line 29: If the previous request is accepted, can we change this message to `"Kicking off thread i of n"`

src/cpp/logger.cpp
```diff
@@ -0,0 +1,240 @@
...
+    else {
+        _log_level = LLInfo;
+    }
+
+    // Now that everything is configured, issue warning and
+    // error messages
+    if (missingLogFile) {
+        log_warning(
+            LLInfo,
+            "Environment variable SS_LOG_FILE is not set. "
```
MattToast, line 119: typo: SS -> SR

src/cpp/logger.cpp
```diff
@@ -0,0 +1,240 @@
...
+            "Defaulting to stdout"
+        );
+    }
+    if (missingLogLevel) {
+        log_warning(
+            LLInfo,
+            "Environment variable SS_LOG_LEVEL is not set. "
```
MattToast, line 126: typo: SS -> SR

MattToast (commented):
A couple of small requests that I have after parsing some of the logs from the scaling study.

src/cpp/logger.cpp
billschereriii: Fixed, thanks for the catch

src/cpp/logger.cpp
billschereriii: Fixed, thanks for the catch

src/cpp/threadpool.cpp
billschereriii: see previous

mellis13 (approved):
Thanks for working so hard on this logging functionality!   Just one small comment/question. Otherwise LGTM

src/fortran/client.F90
```diff
@@ -232,14 +234,30 @@ function SR_error_parser(self, response_code) result(is_error)
...
+  ! Local variables
+  character(kind=c_char, len=:), allocatable :: c_logger_name
+  integer(kind=c_size_t) :: logger_name_length
+
+  if (present(logger_name)) then
+    allocate(character(kind=c_char, len=len_trim(logger_name)) :: c_logger_name)
+    c_logger_name = logger_name
+    logger_name_length = len_trim(logger_name)
+  else
+    allocate(character(kind=c_char, len=10) :: c_logger_name)
```
ashao, line 252: Doesn't really matter, but the length here should be 7

src/fortran/client.F90
```diff
@@ -232,14 +234,30 @@ function SR_error_parser(self, response_code) result(is_error)
...
+    c_logger_name = 'default'
+    logger_name_length = 8
```
ashao, line 254: More important here

ashao (approved):
Thanks for all the extra work on this PR. This logging functionality is a huge quality of life and productivity boost for application developers.

src/python/bindings/bind.cpp
```diff
@@ -26,8 +26,14 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "pysrobject.h"
 #include "pyclient.h"
+#include "pydataset.h"
+#include "pylogcontext.h"
 #include "srexception.h"
+#include "logger.h"
+//#include "srobject.h"
```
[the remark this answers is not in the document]
billschereriii: I'll clean this up as part of ticket #338. It doesn't really hurt anything before then.

src/fortran/client.F90
billschereriii: used len_trim to eliminate hard-coded lengths

src/fortran/client.F90
billschereriii: used len_trim to eliminate hard-coded lengths

src/python/module/smartredis/logcontext.py
```diff
@@ -0,0 +1,79 @@
...
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from .smartredisPy import PyLogContext
+from .util import exception_handler, typecheck
+
+from .error import *
```
MattToast, line 30: Nothing from error is used

src/python/module/smartredis/srobject.py
```diff
@@ -0,0 +1,79 @@
...
+
+class SRObject:
+    def __init__(self, context):
+        """Initialize a SRObject object
+
+        :param context: logging context
+        :type name: str
+        """
+        typecheck(context, "context", str)
+        self._name = context
```
MattToast, line 40: Storing the ctx in a "protected" variable, but doesn't seem to be used internally within the class. Could be problematic if an app dev were to instance and then call `set_data`. Maybe remove or replace with
```py
@property
@exception_handler
def name(self):
    # Note: This would require adding new bindings
    self._data.get().get_context()
```

MattToast (approved):
LGTM!! Couple of last minute recommendations for the python side of things, but overall everything looks fantastic! Thanks for the hard work on this one!!

src/python/module/smartredis/srobject.py
billschereriii: Object-based context in Python isn't working yet, but will be resolved shortly as part of #338

src/python/module/smartredis/logcontext.py
```diff
@@ -0,0 +1,79 @@
...
+
+class LogContext:
+    def __init__(self, context):
+        """Initialize a LogContext object
+
+        :param context: logging context
+        :type name: str
+        """
+        typecheck(context, "context", str)
+        self._name = context
```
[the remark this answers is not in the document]
billschereriii: Object-based context in Python isn't working yet, but will be resolved shortly as part of #338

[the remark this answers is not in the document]
billschereriii: But the property approach looks completely viable

PR #282, opened 2022-11-17
Change dependency for build-examples

ashao:
`build-examples` was pointing to the `lib` target which does not build the Fortran client library. This causes the post-merge tests to fail. Updating this to `lib-with-fortran` allows compilation of the example codes to succeed.


mellis13 (approved):
LGTM!

PR #285, opened 2022-12-15
Make PyClient a property of Python Client class, not the parent

billschereriii:
This enables future changes in support of Python object-context logging


MattToast (approved):
LGTM!! Thanks for fixing this discrepancy

PR #293, opened 2023-01-24
Add serialization support for Client and Dataset 

billschereriii:
C, C++, Fortran get to_string() methods
C++ also gets ostream& operator<<() methods
Python gets __str__() methods

Sample output (Client):

```python
>>> from smartredis import Client
>>> c = Client("127.0.0.1:6381", cluster=True, logger_name="testcase"); print(c)
Client (testcase):
Clustered Redis connection:
  Redis shards at:
    127.0.0.1:6383
    127.0.0.1:6381
    127.0.0.1:6382
  Protocol: TCP
  Command parameters:
    Retry attempts: 101
    Retry interval (ms): 1000
    Attempt timeout (ms): 100
  Connection parameters:
    Retry attempts: 101
    Retry interval (ms): 1000
    Attempt timeout (ms): 100
  Threadpool: 4 threads
```

Sample output (DataSet):

```python
>>> from smartredis import Dataset
>>> import numpy as np
>>> d = Dataset("This is a test")
>>> data = np.uint8([2,4,8])
>>> d.add_tensor("u8_tensor", data)
>>> data = np.double([2.0,4.1,8.3, 5.6])
>>> d.add_tensor("double_tensor", data)
>>> d.add_meta_scalar("float2_scalar", np.float(3.1415926535))
>>> d.add_meta_scalar("float_scalar", np.double(3.1415926535))
>>> d.add_meta_string("metastring", "metavalue")
>>> print(d)
DataSet (This is a test):
Tensors:
  double_tensor:
    type: double
    dimensions: [4]
    elements: 4
  u8_tensor:
    type: 8 bit unsigned integer
    dimensions: [3]
    elements: 3
Metadata:
  metastring:
    type: string
  float2_scalar:
    type: double
  float_scalar:
    type: double
```


src/python/bindings/bind.cpp
```diff
@@ -54,6 +52,7 @@ PYBIND11_MODULE(smartredisPy, m) {
         .def(py::init<std::string&>());
 
     // Python client class
+    #define CLIENT_METHOD(name)  def(#name, &PyClient::name)
```
MattToast, line 55: Is this macro used? (Like the idea of using it)

src/python/bindings/bind.cpp
```diff
@@ -112,9 +111,11 @@ PYBIND11_MODULE(smartredisPy, m) {
         .def("poll_list_length_gte", &PyClient::poll_list_length_gte)
         .def("poll_list_length_lte", &PyClient::poll_list_length_lte)
         .def("get_datasets_from_list", &PyClient::get_datasets_from_list)
-        .def("get_dataset_list_range", &PyClient::get_dataset_list_range);
+        .def("get_dataset_list_range", &PyClient::get_dataset_list_range)
+        .def("to_string", &PyClient::to_string);
 
     // Python Dataset class
+    #define DATASET_METHOD(name) def(#name, &PyClient::name)
```
MattToast, line 118: Same "macro used" question here

tests/python/test_dataset_methods.py
```diff
@@ -27,6 +27,19 @@
...
+    """
+    dataset = Dataset("test-dataset")
+    data = np.uint8([2,4,8])
+    dataset.add_tensor("u8_tensor", data)
+    data = np.double([2.0,4.1,8.3, 5.6])
+    dataset.add_tensor("double_tensor", data)
+    dataset.add_meta_scalar("float2_scalar", float(3.1415926535))
+    dataset.add_meta_scalar("float_scalar", np.double(3.1415926535))
+    dataset.add_meta_string("metastring", "metavalue")
+    print(dataset)
```
MattToast, line 41: Little worried about this test: technically speaking, w/o the changes in this PR, this should still pass.

Maybe just check to make sure that the string of a dataset is no equal to the repr (which is the default when casting an object to a str)

```py
dataset = Dataset(...)
...
assert repr(dataset) != str(dataset)
```

Or you could search for a sub-string that we expect to be in the "string-ified" dataset

```py
assert "Dataset(test-dataset):" in str(dataset)
```

tests/python/test_address.py
```diff
@@ -28,6 +28,12 @@
 
 from smartredis import Client
 
+def test_serialization(use_cluster, context):
+    # get env var to set through client init
+    ssdb = os.environ["SSDB"]
+    c = Client(address=ssdb, cluster=use_cluster, logger_name=context)
+    print(c)
```
MattToast, line 35: Similar worry to dataset test: this test should pass reguardless of changes made in this PR. Maybe do the same `repr(c) != str(c)` trick?

MattToast (changes requested):
Looked through the `*.py/pybind/Py*.cpp` files in detail and have a couple of small notes, but otherwise everything looks good. Skimmed through the reset and didn't see any immediate red flags.

Overall looks great!! This should be very helpful for users attempting to debug their simulations!!

src/python/bindings/bind.cpp
billschereriii: Not yet -- it will be a separate PR to switch over to it

src/python/bindings/bind.cpp
billschereriii: Not yet -- it will be a separate PR to switch over to it

tests/fortran/client_test_dataset.F90
```diff
@@ -195,6 +196,11 @@ program main
   if (result .ne. SRNoError) error stop
   if (.not. all(meta_int64_recv == meta_int64_vec)) error stop 'meta_int64: FAILED'
 
+  ! Test dataset serialization
+  write (*,*) dataset%to_string()
```
ashao, line 200: Might be better to compare this to what the expected string should be just to make sure `to_string` isn't just returning garbage

tests/python/test_dataset_methods.py
billschereriii: Done

tests/python/test_address.py
billschereriii: Done

src/fortran/client/client_interfaces.inc
```diff
@@ -152,3 +152,11 @@ interface
     logical(kind=c_bool)          :: exists
   end function poll_key_c
 end interface
+
+interface
+  function client_to_string_c(client) bind(c, name="client_to_string")
+    use iso_c_binding, only : c_ptr, c_char
+    type(c_ptr)                   :: client_to_string_c
```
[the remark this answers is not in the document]
billschereriii: I think we tried that last time (on the get_last_error() PR) and it didn't work

tests/fortran/client_test_dataset.F90
```diff
@@ -195,6 +196,11 @@ program main
...
+  call dataset%print_dataset()
+  call dataset%print_dataset(stderr)
```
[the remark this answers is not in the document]
billschereriii: Will do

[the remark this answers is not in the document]
billschereriii: Done

tests/fortran/client_test_dataset.F90
billschereriii: Done

MattToast (approved):
LGTM!!

src/cpp/dataset.cpp
```diff
@@ -367,3 +370,41 @@ TensorBase* DataSet::_get_tensorbase_obj(const std::string& name)
...
 }
+
+// Create a string representation of the DataSet
+std::string DataSet::to_string() const
+{
+    std::string result;
+    result = "DataSet (" + _lname + "):\n";
+
+    // Tensors
+    result += "Tensors:\n";
```
mellis13, line 381: Very minor, but if number or tensors or number of metadata fields is 0, should we print something like ``None`` under the heading:

```
Metadata:
  None
```

So the message looks more complete when one is empty?

mellis13 (approved):
Just one minor question/suggestion.  Otherwise, great addition to the library and LGTM!

src/cpp/dataset.cpp
billschereriii: That's a great suggestion!

PR #297, opened 2023-01-28
Clean up pybind file

billschereriii:
Use macros to simplify client and dataset bindings


MattToast (approved):
New macros look great and really help clean up code intent! LGMT pending tests!!

PR #299, opened 2023-02-02
Add dataset inspection of tensor dimensions

billschereriii:
Implement new dataset get_tensor_dims() method in all four clients and add test support

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -58,7 +58,7 @@ Detailed Notes
 
 .. _PR301: https://github.com/CrayLabs/SmartRedis/pull/301
 .. _PR300: https://github.com/CrayLabs/SmartRedis/pull/300
-.. _PR296: https://github.com/CrayLabs/SmartRedis/pull/299
+.. _PR299: https://github.com/CrayLabs/SmartRedis/pull/299
 .. _PR298: https://github.com/CrayLabs/SmartRedis/pull/298
 .. _PR295: https://github.com/CrayLabs/SmartRedis/pull/295
 .. _PR294: https://github.com/CrayLabs/SmartRedis/pull/294
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ This section details changes made in the development branch that have not yet be
 
 Description
 
+- Add support to inspect the dimensions of a tensor via get_tensor_dims()
 - Update to the latest version of PyBind
 - Change documentation theme to sphinx_book_theme and fix doc strings
 - Add print capability for Client and DataSet
@@ -29,6 +30,7 @@ Description
 
 Detailed Notes
 
+- Added support to inspect the dimensions of a tensor via new get_tensor_dims() method (PR296_)
 - Updated from PyBind v2.6.2 to v2.10.3 (PR295_)
 - Change documentation theme to sphinx_book_theme to match SmartSim documentation theme and fix Python API doc string errors (PR294_)
 - Added print capability for Client and DataSet to give details diagnostic information for debugging (PR293_)
@@ -48,6 +50,7 @@ Detailed Notes
 - Implemented support for Unix Domain Sockets, including refactorization of server address code, test cases, and check-in tests. (PR252_)
 - A new make target `make lib-with-fortran` now compiles the Fortran client and dataset into its own library which applications can link against (PR245_)
 
+.. _PR296: https://github.com/CrayLabs/SmartRedis/pull/296
 .. _PR295: https://github.com/CrayLabs/SmartRedis/pull/295
 .. _PR294: https://github.com/CrayLabs/SmartRedis/pull/294
 .. _PR294: https://github.com/CrayLabs/SmartRedis/pull/293
--- a/include/c_dataset.h
+++ b/include/c_dataset.h
@@ -248,6 +248,25 @@ SRError get_tensor_type(
     void* dataset, const char* name, size_t name_len, SRTensorType* ttype);
 
 /*!
+*   \brief Retrieve the dimensions of a Tensor in the DataSet
+*   \param dataset The dataset to use for this operation
+*   \param name The name of the tensor (null-terminated string)
+*   \param name_len The length in bytes of the tensor name
+*   \param dims A buffer in which to receive the dimensions for
+*               the specified tensor.
+*   \param ndims The size of the buffer supplied in \p dims when this function
+*                is called; receives the number of dimensions supplied on exit.
+*                If the supplied the buffer isn't large enough to contain all
+*                the tensor dimensions, \p ndims will still receive the needed
+*                size.
+*   \return Returns SRNoError on success or an error code on failure. If the
+*           buffer supplied in \p dims is too small, returns SRBadAllocError.
+*/
+SRError get_tensor_dims(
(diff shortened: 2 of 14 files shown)


include/c_dataset.h
```diff
@@ -247,6 +247,25 @@ SRError get_tensor_names(
...
     void* dataset, const char* name, size_t name_len, SRTensorType* ttype);
 
+/*!
+*   \brief Retrieve the dimensions of a Tensor in the DataSet
+*   \param dataset The dataset to use for this operation
+*   \param name The name of the tensor (null-terminated string)
+*   \param name_len The length in bytes of the tensor name
+*   \param dims A buffer in which to receive the dimensions for
+*               the specified tensor.
+*   \param ndims The size of the buffer supplied in \p dims when this function
```
mellis13, line 257: The behavior of this function seems different than other functions like ``get_tensor_names``.  This function assumes you know the dimensions to be able to return the dims?  Is that because we don't want to handle the memory or is there another reason having to do with Fortran?  Would it make sense just to return ``dims`` for the user so they don't have to worry about it and get ``SRBadAllocError``?

mellis13 (changes requested):
Thanks for adding this feature.  Just 1 inline comment/question.  I'm also wondering if we should have C++ unit tests for successful and errors in ``get_tensor_dims``?  The Fortran implicitly does that, but we don't get code coverage from Fortran.

include/c_dataset.h
billschereriii: Exactly so -- it makes sure that we can avoid leaking memory by having the caller supply the buffer into which we copy the dims. If we just returned the dims, we'd have to allocate memory and then give the caller an interface through which to say they don't need this block any more -- which gets really messy really fast. With this approach, if the caller doesn't provide a big enough buffer, we can tell them how big the buffer needs to be, so worst case, they'll only have to call the method twice.

billschereriii:
Skipping C++ unit tests was definitely an oversight. I will get them added before I resubmit this one

mellis13 (approved):
LGTM -- thanks!

src/fortran/dataset.F90
```diff
@@ -478,6 +480,33 @@ function get_tensor_type(self, name, ttype) result(code)
   code = get_tensor_type_c(self%dataset_ptr, c_name, name_length, ttype)
 end function get_tensor_type
 
+
+!> Retrieve the dimensions for a tensor into a supplied buffer, or receive the
+!! number of dimensions if the supplied buffer is too small
+function get_tensor_dims(self, name, dims, dims_length) result(code)
```
ashao, line 486: For this one, I'm half wondering whether we want to be this strongly prescriptive about the `kind` of dims and dims_length. It would require an extra copy for both these variables, but they're so small that I can't think the hit would be significant.

So in the subroutine declaration it would just be:
```
integer, dimension(:), intent(inout) :: dims
integer,               intent(inout) :: dims_length

! Local variables
integer(kind=c_size_t), dimension(size(dims)), target :: c_dims
integer(kind=c_size_t) ::  c_dims_length
...
dims = c_dims
dims_length = c_dims_length

File: src/fortran/dataset.F90 [hunk 3]
```diff
@@ -478,6 +480,33 @@ function get_tensor_type(self, name, ttype) result(code)
...
+  ! local variables
+  character(kind=c_char, len=len_trim(name)) :: c_name
+  integer(kind=c_size_t) :: name_length
+  type(c_ptr) :: dims_ptr
+
+  c_name = trim(name)
+  name_length = len_trim(c_name)
+
+  if (dims_length .gt. size(dims)) then
+    error stop 'undersized buffer in call to get_tensor_dims'
```
ashao, line 502: Add the minimum size of `dims` that the user would need to specify to the error message

ashao (changes requested):
One comment about specifically requiring that the user provides a `kind=c_size_t` buffer.

src/fortran/dataset.F90
ashao: Oh teh other thing about this, it follows the pattern of the rest of the Fortran SmartRedis client where we're trying NOT to make the user know about the `iso_c_binding` stuff in their own code.

src/python/module/smartredis/dataset.py
```diff
@@ -229,3 +229,13 @@ def get_tensor_names(self):
         :rtype: list
         """
         return self._data.get_tensor_names()
+
+    @exception_handler
+    def get_tensor_dims(self, name):
+        """Get the dimensions of a tensor in the DataSet
+
+        :return: a list of the tensor dimensions
+        :rtype: list
```
MattToast, line 238: incredibly minor doc request: can we change the return type to `list[int]`

doc/changelog.rst
```diff
@@ -54,6 +56,7 @@ Detailed Notes
 
 .. _PR301: https://github.com/CrayLabs/SmartRedis/pull/301
 .. _PR300: https://github.com/CrayLabs/SmartRedis/pull/300
+.. _PR296: https://github.com/CrayLabs/SmartRedis/pull/296
```
MattToast, line 59: typo: Pull request 299?

MattToast (changes requested):
Took a look at the c/c++/py code changes. A couple of small doc related changes, but other than that and pending checking in of python tests everything here looks great!!

src/fortran/dataset.F90
billschereriii: Sounds completely reasonable and more user-friendly

src/fortran/dataset.F90
billschereriii: We don't know it yet. This is the case where they pass us a buffer for the dims, say of size 3 and then tell us it's bigger than that, say size 10. If I ignore that, we risk having memory corruption and a segfault, so I want to warn them straightaway. I could just silently truncate down num_dims before proceeding but I'd rather make the caller aware of the error they've made.

doc/changelog.rst
billschereriii: I promise I'll get the changelog sorted out before the final commit of these two PRs

src/python/module/smartredis/dataset.py
billschereriii: done, and also changed the rtype for the other lists to list[str]

src/fortran/dataset.F90
billschereriii: Made the error message clearer

doc/changelog.rst
```diff
@@ -56,6 +58,7 @@ Detailed Notes
 
 .. _PR301: https://github.com/CrayLabs/SmartRedis/pull/301
 .. _PR300: https://github.com/CrayLabs/SmartRedis/pull/300
+.. _PR296: https://github.com/CrayLabs/SmartRedis/pull/299
```
MattToast, line 61:
```
.. _PR296: https://github.com/CrayLabs/SmartRedis/pull/299
        ^ Should be 9
```
One last typo

MattToast (approved):
LGTM pending one last typo and c/py tests!!

doc/changelog.rst
billschereriii: *sigh*

ashao (approved):
Thanks for addressing my concerns!

PR #300, opened 2023-02-03
Add some missing error handling

billschereriii:
Check the CommandResponse objects returned from set_script() and set_model() calls and throw an internal error if they come back in bad state.

### DIFF
--- a/src/cpp/client.cpp
+++ b/src/cpp/client.cpp
@@ -591,9 +591,14 @@ void Client::set_model(const std::string& name,
     }
 
     std::string key = _build_model_key(name, false);
-    _redis_server->set_model(key, model, backend, device,
-                             batch_size, min_batch_size,
-                             tag, inputs, outputs);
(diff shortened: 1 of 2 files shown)


src/cpp/client.cpp
```diff
@@ -730,7 +735,11 @@ void Client::set_script(const std::string& name,
     }
 
     std::string key = _build_model_key(name, false);
-    _redis_server->set_script(key, device, script);
+    auto response = _redis_server->set_script(key, device, script);
+    if (response.has_error()) {
+        throw SRInternalException(
+            "An unknown error occurred while setting the scriot");
```
MattToast, line 741: typo: script

MattToast (approved):
LGTM, pending one typo!!

src/cpp/client.cpp
billschereriii: fixed

PR #302, opened 2023-02-16
Make Xarray an Optional Dependency

MattToast:
This PR makes the following changes
- Removes `xarray` from SR's `install_requires` list
- Slightly refactors check if xarray is available by converting it to a decorator
- Adds test to make sure the decorator works as expected if `xarray` cannot be imported

### DIFF
--- a/src/python/module/smartredis/dataset_utils.py
+++ b/src/python/module/smartredis/dataset_utils.py
@@ -76,7 +76,6 @@ def typecheck_stringlist(names, strings_name, string_name):
 
 class DatasetConverter:
     @staticmethod
-    @_requires_xarray
     def add_metadata_for_xarray(
         dataset, data_names, dim_names, coord_names=None, attr_names=None
     ):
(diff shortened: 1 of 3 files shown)


billschereriii (approved):
LGTM!  Thank you so much for giving us an example of how to properly set up support for a converter while keeping it optional

src/python/module/smartredis/dataset_utils.py
```diff
@@ -50,6 +76,7 @@ def typecheck_stringlist(names, strings_name, string_name):
 
 class DatasetConverter:
     @staticmethod
+    @_requires_xarray
```
ashao, line 79: This shouldn't need `xarray` as it's all SmartRedis natives.

ashao (approved):
One minor comment where the `@_requires_xarray` doesn't need to be included

src/python/module/smartredis/dataset_utils.py
MattToast: Bring in @billschereriii on this, as this was a recommendation from him

MattToast: reverting

PR #305, opened 2023-02-27
Fix assorted documentation errors and omissions

billschereriii:
Fix assorted documentation errors and omissions


ashao (approved):
Thanks again for chasing these down!

PR #306, opened 2023-02-27
Fix list key prefix

al-rigazzi:
An erroneous `on_db` flag is causing wrong behavior when lists are on the DB. This PR fixes it with a minimum set of changes.


billschereriii (approved):
LGTM! Thank you for taking care of this and for validating that this call site was the only one affected

PR #309, opened 2023-03-06
Pipeline fixes

billschereriii:
Fix pipeline execution to no longer apply a retry timeout before returning,  allow retries to actually occur in the event of a non-fatal error, and to eliminate an unreachable return statement


mellis13 (approved):
Thanks for fixing this!  LGTM

PR #310, opened 2023-03-07
Fix multigpu set_model

al-rigazzi:
This PR fixes what seems to be a bug in setting the model for multiple GPUs.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ This section details changes made in the development branch that have not yet be
 
 Description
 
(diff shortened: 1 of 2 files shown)


ashao (approved):
Nice catch!

billschereriii (approved):
Thanks for catching this! I confirmed that the other 3 call sites (here and in RedisCluster.cpp) were correct already.

billschereriii:
Please update the change log!

al-rigazzi:
Updated - and fixed - the changelog!

PR #312, opened 2023-03-10
Template-based decorators for C layer

billschereriii:
Update the C interface with a template-based pseudo-decorator to elminate much of the boilerplate exception management code and streamline implementation


MattToast (commented):
Absolutely love this!! So much more readable!

Totally has my approval. Simply withholding to give lead reviewer @ashao time to look over

ankona:
> 

It does look like a huge improvement, but I can't help but wonder if you've exposed a different problem - why do we have to instantiate a client 55x in one file (e.g. MAKE_CLIENT_API is called 55x). 


What if you add the client argument to `c_client_api` - you would immediately remove another 55 LoC by doing the `reinterpret_cast` out of the client method code. 

And, I think, since it's also repeated, you could shift the line `std::string name_str(name, name_length);` too, to knock off another 28 LoC. This same repetition pattern seems to repeat in the other clients. 

A final note.. I think the `MAKE_CLIENT_API` method could probably be named slightly better to fit it's actual function or it could be modified to do that one job & return the client without also executing the operation, e.g. `MAKE_CLIENT_API(c_client).execute({ ... })`

billschereriii:
> > 
> 
> It does look like a huge improvement, but I can't help but wonder if you've exposed a different problem - why do we have to instantiate a client 55x in one file (e.g. MAKE_CLIENT_API is called 55x). I _think_ this might be to maintain the same external API for clients, vs just having them pass in the right client type, so it could be ignored due to necessity.
> 
> What if you add the client argument to `c_client_api` - you would immediately remove another 55 LoC by doing the `reinterpret_cast` out of the client method code.
> 
> And, I think, since it's also repeated, you could shift the line `std::string name_str(name, name_length);` too, to knock off another 28 LoC. This same repetition pattern seems to repeat in the other clients.
> 
> A final note.. I think the `MAKE_CLIENT_API` method could probably be named slightly better to fit it's actual function or it could be modified to do that one job & return the client without also executing the operation, e.g. `MAKE_CLIENT_API(c_client).execute({ ... })`

Per offline conversation, opening a new issue to examine moving reinterpret_casts inside the lambdas. Not immediately clear how to automate processing of string in the lambdas -- likely to be more hassle than it's worth, but will keep an eye out. Naming of macro will stay unchanged to keep consistency with PyBind layer.

ashao (approved):
This is a really innovative solution that really makes the client base a lot more extensible and maintainable. @billschereriii has allayed all concerns that I had which included understanding how to handle specific exceptions. The decorator-approach in its current manifestation is flexible enough to handle the main edge cases that I could think of.

Great work Bill!

src/c/c_client.cpp
```diff
@@ -272,33 +211,21 @@ SRError put_tensor(void* c_client,
...
-                  const char* name,
-                  const size_t name_length,
-                  void** result,
-                  size_t** dims,
-                  size_t* n_dims,
-                  SRTensorType* type,
-                  const SRMemoryLayout mem_layout)
+extern "C" SRError get_tensor(
+  void* c_client,
+    const char* name,
```
MattToast, line 220: pedantic style thing: can we have all these args at the same col?

MattToast (approved):
lgtm!! Thanks for this, this is massive clean up!!

PR #317, opened 2023-03-22
Prevent codecov status checks from failing

ashao:
We have wide variability of our total coverage reports even for non-substantive commits. We suspect that this is because we are compiling unit tests and the SmartRedis library in cmake's Release mode which enables -O3 and may, at worse, skip the execution of some tests and at least confuses the tracer that marks which lines have been run.

As a workaround for now, we are lowering the target thresholds to 0% so that commits are not marked with the dreaded 'X' due to the variance that results in code coverage.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,8 +5,10 @@ coverage:
   status:
     project:
       default:
-        target: 60%
-        threshold: 2%
+        target: 0%
+    patch:
+      default:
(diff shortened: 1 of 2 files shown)


billschereriii (approved):
LGTM! Thanks, and hopefully we can get these spurious failures eliminated in the next release cycle

PR #320, opened 2023-03-24
Fix copy/paste error in to_str methods in C interface

billschereriii:
boilerplate cloning issue


MattToast (approved):
Good catch! LGTM pending tests!!

PR #330, opened 2023-04-24
Add pip-install target to Makefile

billschereriii:
square bracket options, notably [dev], aren't supported as they won't be needed once the build refactor is complete.

Test via: 

1. pip uninstall smartredis
2. make pip-install
3. make pip-install


MattToast (approved):
LGTM!!

PR #331, opened 2023-04-27
third-party software support in Makefile

billschereriii:
Third-party software is now downloaded, built, and installed directly from the Makefile, and deps/test-deps are now properly listed as dependencies for lib and testing targets, respectively.

I've left the old scripts intact for now so that they are available for reference.

Two weirdnesses I noticed in testing:
1. `Make clean` nukes the installed versions of hiredis and redis++. I couldn't come up with a good way to stop that from happening. so as a stopgap, I nuke the third-party built versions before downloading, building, installing them
2. From an empty state (after `make clobber`) the sequence `make deps` => `make test-deps` reports a failure at the very end of installing RedisAI even though everything has successfully installed. Runing `make test-deps` again finishes the installation. This doesn't happen with any of the following sequences:

   - `make clobber` => `make test-deps` => `make deps`
   - `make clobber` => `make deps` => `make test` (which invokes `test-deps` first thing)

From a fresh clone, `make test` will now install everything and run the tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -366,7 +366,7 @@ third-party/RedisAI/install-cpu/redisai.so:
 	@mkdir -p third-party
 	@cd third-party && \
 	GIT_LFS_SKIP_SMUDGE=1 git clone --recursive $(REDISAI_URL) RedisAI --branch $(REDISAI_VER) --depth=1
-	@cd third-party/RedisAI && \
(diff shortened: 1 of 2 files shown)


billschereriii:
As an aside, I expect that at least some of the weirdnesses above are reproducible with the existing build system

PR #334, opened 2023-04-29
Update redis++ to 1.3.8

### LINKED ISSUE #333: redis-plus-plus pkgconfig doesn't match actual install location
# Description
When installing redis++ through the normal build instructions, a `redis++.pc` file is created. Sometimes, CMake decides to install it's libraries into `lib64` rather than the usual `lib` directory, but the `redis++.pc` file will still point to the `lib` directory. Why it decides that, I don't know. But I've run into it.

I've submitted an issue to https://github.com/sewenew/redis-plus-plus/issues/474 (which also contains more info about the above issue). This has now been fixed, but to have that apply for smartredis installs, the redis++ version in this repo needs to be updated.

As seen in the above-linked issue, I've requested that redis++ have a new release to include the fix so y'all can simply change the version associated with the redis-plus-plus install. I wanted to go ahead and submit an issue on it before I forget, or if y'all are willing to pin the redis++ version to the [current master commit](https://github.com/sewenew/redis-plus-plus/commit/f1c02ee07c7b81295c021ef1236c646113174fdc) at redis++.

jrwrigh:
Resolves #333

### DIFF
--- a/build-scripts/build_deps.sh
+++ b/build-scripts/build_deps.sh
@@ -40,7 +40,7 @@ if ls ../install/lib/libredis++.a 1>/dev/null 2>&1; then
     echo "Redis-plus-plus has already been installed"
 else
     if [[ ! -d "./redis-plus-plus" ]]; then
-        git clone https://github.com/sewenew/redis-plus-plus.git redis-plus-plus --branch 1.3.5 --depth=1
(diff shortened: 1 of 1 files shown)


ashao:
Thanks @jrwrigh! We really appreciate the time you took to track this down

ashao (approved):
Thanks again!

PR #339, opened 2023-05-09
Switch CI/CD pipeline to use MacOS 12 instead of 10.15

billschereriii:
The MacOS 10.15 runner has been deprecated for a while and is now kaput


ashao (approved):
Thanks for updating this. Approved pending succesfful completion of tests

PR #340, opened 2023-05-09
Refactor to return default Fortran logical kind

### LINKED ISSUE #337: Fortran `poll_tensor` logical type of `exists`
The `poll_tensor` implementation for Fortran requires a `logical(kind=c_bool)` type for the `exists` flag. Why is that? Wouldn't a normal `logical` work? Or, otherwise, couldn't this be converted within the function? It seems a bit confusing that users have to use `iso_c_binding` on their side. Also the [Fortran API docs](https://www.craylabs.org/docs/api/smartredis_api.html#f/smartredis_client/poll_tensor) does not mention it.

ashao:
Many routines were returning logical(kind=c_bool) which is not the same default kind used for logicals by some compilers (e.g. GCC sets logical to be kind=4). Now these functions return the default logical kind so that users will not have to import iso_c_binding into their code.

Closes #337


src/fortran/configoptions.F90
```diff
@@ -15,7 +15,7 @@
 !
 ! THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 ! AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-! IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+! IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AR
```
billschereriii, line 18: Good catch!

billschereriii (approved):
LGTM, thanks for taking care of this one!

PR #346, opened 2023-05-25
Downgrade F2018 requirement to F2003 (NVHPC support)

ashao:
Replaces the assumed rank feature of F2018 used in the Fortran client with assumed
shape arrays. Compilers need only be compliant with the F2003 standard now. While
this means that it is possible to compile SmartRedis with the Nvidia toolchain,
users should consider this only experimental support until we can incorporate
Nvidia compilers into our CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -389,7 +389,7 @@ install/lib/libredis++.a:
 	cmake -DCMAKE_BUILD_TYPE=Release -DREDIS_PLUS_PLUS_BUILD_TEST=OFF \
 		-DREDIS_PLUS_PLUS_BUILD_SHARED=OFF -DCMAKE_PREFIX_PATH="../../../install/lib/" \
 		-DCMAKE_INSTALL_PREFIX="../../../install" -DCMAKE_CXX_STANDARD=17 \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -309,9 +309,8 @@ install/lib/libredis++.a:
 	@cd third-party/redis-plus-plus && \
 	mkdir -p compile && \
 	cd compile && \
-	cmake -DCMAKE_BUILD_TYPE=Release -DREDIS_PLUS_PLUS_BUILD_TEST=OFF -DREDIS_PLUS_PLUS_BUILD_SHARED=OFF -DCMAKE_PREFIX_PATH="../../../install/lib/" -DCMAKE_INSTALL_PREFIX="../../../install" -DCMAKE_CXX_STANDARD=17 .. && \
-	CC=gcc CXX=g++ make -j && \
```
billschereriii, line 313: Does deleting this line work with the other compilers?

billschereriii:
Approval is conditional of course on test cases passing

Makefile
ashao: Yes, I think the intention was always to use the GCC compilers (see old line 313).

PR #349, opened 2023-06-02
Parallelize build of examples

billschereriii:
Enable parallel build of examples by putting Fortran utility files into a mini-library


ashao (approved):
Thanks for this @billschereriii

PR #352, opened 2023-06-08
Remove debug statement

mellis13:
This PR removes a debug statement that was left in the pybind layer.  This statement is surfaced whenever the user invokes ``put_dataset``.


billschereriii (approved):
That looks like my handiwork, unfortunately. Thanks for cleaning it up

PR #356, opened 2023-06-22
SmartRedis test system refactor

billschereriii:
Massive overhaul of test system, automating launch and teardown of Redis and removing the need for the setup_test_env.sh script


Makefile
```diff
@@ -278,90 +299,187 @@ endif
...
+# 	1: the test directory in which to run tests
+define run_smartredis_tests_with_standalone_server
+	echo "Launching standalone Redis server" && \
+	export SR_TEST_DEVICE=$(SR_TEST_DEVICE) SR_SERVER_MODE=Standalone && \
+	export SMARTREDIS_TEST_CLUSTER=False SMARTREDIS_TEST_DEVICE=$(SR_TEST_DEVICE) && \
+	export SSDB=127.0.0.1:$(SR_TEST_PORT) && \
+	python utils/launch_redis.py --port $(SR_TEST_PORT) --nodes 1 \
+		--rai $(SR_TEST_REDISAI_VER) --device $(SR_TEST_DEVICE) && \
+	echo "Running standalone tests" && \
+	PYTHONFAULTHANDLER=1 python -m pytest $(SR_TEST_PYTEST_FLAGS) $(COV_FLAGS) \
```
MattToast, line 321: Incredibly obscure corner case that I'm not sure is a problem or that we even want to handle, but thought I would bring to your attention just in case:

If a new dev, in a brand new python env, runs `make test` but does not set `SR_PYTHON=ON`, then `pip install -e .[dev]` will never get run, and thus `pytest` will never be installed, and thus this command will error out here and leave the a redis db running.

(like I said, very obscure, not even sure its an issue, hahaha)

MattToast: A similar error can occur for the clustered/uds version of this fn

MattToast (changes requested):
Looks amazing!

I went on a deep dive into the `launch_redis.py` file basically just trying to remove the need for `shell=True` in the `Popen` calls and calling out some style things along the way.

I skimmed through the `Makefile` as well and that, to the best of my ability looks great with one outstanding questiong.

Overall, only some pedantic feedback! LGTM!!

utils/launch_redis.py
```diff
@@ -0,0 +1,255 @@
...
+import pathlib
+
+def check_availability(n_nodes, port, udsport):
+    """Repeat a command until it is successful
+    """
+    num_tries = 5
+    is_uds = udsport is not None
+    if is_uds:
+        n_nodes = 1
+    cicd = os.getenv('SR_CICD_EXECUTION')
```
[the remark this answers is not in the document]
billschereriii: I think we want to control our own destiny, so I don't want to use a built-in. I can probably be talked out of that stance though

utils/launch_redis.py
```diff
@@ -0,0 +1,255 @@
...
+        rediscli = os.path.abspath(
+            os.path.dirname(__file__) + "/../third-party/redis/src/redis-cli"
+        )
+
+    # Clobber the server(s)
+    procs = []
+    for i in range(n_nodes):
+        connection = f"-s {udsport}" if is_uds else f"-p {str(port + i)}"
+        cmd = f"{rediscli} {connection} shutdown"
+        print(cmd)
```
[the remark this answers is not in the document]
billschereriii: This was in fact intentional. It helps when reading the test output to see where things went. We can potentially remove it later but I'd like to keep it around for now

Makefile
billschereriii: It's a potential issue for sure, but not one introduced with this PR. The only real way to address it would be to put a call to pip install into the Makefile, and as a group we decided against that with the build system refactor PR

utils/launch_redis.py
```diff
@@ -0,0 +1,255 @@
...
+    # Let exceptions propagate to the caller
+    check_availability(n_nodes, port, udsport)
+    for proc in procs:
+        _,_ = proc.communicate(timeout=15)
+        if proc.returncode != 0:
+            raise RuntimeError("Failed to launch Redis server!")
+
+    # Create cluster for clustered Redis request
+    if n_nodes > 1:
+        cluster_str=' '
```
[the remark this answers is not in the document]
billschereriii: yeah, not needed

utils/launch_redis.py
```diff
@@ -0,0 +1,257 @@
...
+import pathlib
+
+def check_availability(n_nodes, port, udsport):
+    """Repeat a command until it is successful
+    """
+    num_tries = 5
+    is_uds = udsport is not None
+    if is_uds:
+        n_nodes = 1
+    cicd = os.getenv('SR_CICD_EXECUTION')
```
ashao, line 40: I think the easier way to do this is to use `shutil.which('redis-cli')` and then set the `rediscli` variable at the top of the file and/or make it into a small function

ashao (approved):
Looks good to me, with the exception of the one change to the detection of `redis-cli`. This should avoid the need to set another environment variable.

Thanks for this; it's a HUGE improvement

utils/launch_redis.py
billschereriii: We actually have fairly substantial changes based on whether we're in CI/CD or not, so this isn't the only thing affected. Can discuss more offline

utils/launch_redis.py
```diff
@@ -0,0 +1,257 @@
...
+    check_availability(n_nodes, port, udsport)
+    for proc in procs:
+        _ = proc.communicate(timeout=15)
+        if proc.returncode != 0:
+            raise RuntimeError("Failed to launch Redis server!")
+
+    # Create cluster for clustered Redis request
+    if n_nodes > 1:
+        cluster_str = ''
+        cluster_str += " ".join(f"127.0.0.1:{port + i}" for i in range(n_nodes))
```
MattToast, line 232: One last thing to remove an unnecessary assignment/strcpy 
```suggestion
        cluster_str = " ".join(f"127.0.0.1:{port + i}" for i in range(n_nodes))
```

MattToast (approved):
LGTM!! Ship it!!

PR #357, opened 2023-06-22
Fix memory leak in C layer get_string_option()

billschereriii:
Put string buffers in stash that we can clear out at destruction time


MattToast (approved):
LGTM!!

PR #359, opened 2023-06-24
Streamline setup.py with new build interface

billschereriii:


setup.py
```diff
@@ -50,65 +49,72 @@ def cmake(self):
...
+
     def run(self):
+        # Validate dependencies
+        check_prereq("cmake")
         check_prereq("make")
         check_prereq("gcc")
         check_prereq("g++")
 
+        # Set up parameters
+        source_directory = Path(os.path.abspath(os.path.dirname(__file__))).resolve()
```
MattToast, line 66: Super minor, but if you want to make this a hair simpler, this should be the equivalent:
```py
source_directory = Path(__file__).parent.resolve()
```

setup.py
```diff
@@ -50,65 +49,72 @@ def cmake(self):
...
-        cmake_cmd = [self.cmake, '--build', '.'] + self.build_args
-        subprocess.check_call(cmake_cmd,
-                              cwd=build_directory)
-
-        shutil.copytree(setup_path.joinpath("install"),
-                        build_directory.joinpath("install"))
+        # Build dependencies
+        print('-'*10, 'Building third-party dependencies', '-'*40)
+        subprocess.check_call(
+            [f"{self.make}", "deps"],
```
MattToast, line 79: Super minor thing here:
It looks like `self.make` should already be returning a `str | None` and we know that it will not be `None` bc the previous `check_prereq` did not error out, so the `f"{self.make}"` looks like a redundant cast to a string.

MattToast (approved):
Hooray! A bunch of random header files are no longer being copied into the build!!

I have a couple of small suggestions, but nothing worth holding up approval over! LGTM!!

PR #362, opened 2023-06-28
Add Typehints

billschereriii:
Add type hints to the SmartRedis library Python interface

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -9,6 +9,7 @@ To be released at some future point in time
 Description
 
 - Remove debug output in pybind layer
+- Update Hiredis version to 1.1.0
 - Enable parallel build for the SmartRedis examples
 - Experimental support for Nvidia toolchain
 - Major revamp of build and test systems for SmartRedis
@@ -23,6 +24,7 @@ Description
 Detailed Notes
 
 - Remove debug output in pybind layer associated with put_dataset (PR352_)
+- Updated to the latest version of Hiredis (1.1.0) (PR351_)
 - Enable parallel build for the SmartRedis examples by moving utility Fortran code
 into a small static library (PR349_)
 - For the NVidia toolchain only: Replaces the assumed rank feature of F2018 used
@@ -43,6 +45,7 @@ users need not import `iso_c_binding` in their own applications (PR340_)
 - Added ConfigOptions class and API, which will form the backbone of multiDB support (PR303_)
 
 .. _PR352: https://github.com/CrayLabs/SmartRedis/pull/352
+.. _PR351: https://github.com/CrayLabs/SmartRedis/pull/351
 .. _PR349: https://github.com/CrayLabs/SmartRedis/pull/349
 .. _PR346: https://github.com/CrayLabs/SmartRedis/pull/346
 .. _PR341: https://github.com/CrayLabs/SmartRedis/pull/341
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -8,6 +8,7 @@ To be released at some future point in time
 
 Description
 
+- Updated setup.py to work with the new build system
 - Remove unneeded method from Python SRObject class
 - Fixed a memory leak in the C layer
 - Remove debug output in pybind layer
@@ -25,6 +26,7 @@ Description
 
 Detailed Notes
 
+- Streamlined setup.py to simplify Python install (PR359)
 - Remove from_pybind() from Python SRObject class as it's not needed and didn't work properly anyway (PR358_)
 - Fixed memory leaked from the C layer when calling get_string_option() (PR357_)
 - Remove debug output in pybind layer associated with put_dataset (PR352_)
@@ -48,6 +50,7 @@ users need not import `iso_c_binding` in their own applications (PR340_)
 - New pip-install target in Makefile will be a dependency of the lib target going forward so that users don't have to manually pip install SmartRedis in the future (PR330_)
 - Added ConfigOptions class and API, which will form the backbone of multiDB support (PR303_)
 
+.. _PR359: https://github.com/CrayLabs/SmartRedis/pull/359
 .. _PR358: https://github.com/CrayLabs/SmartRedis/pull/358
 .. _PR357: https://github.com/CrayLabs/SmartRedis/pull/357
 .. _PR352: https://github.com/CrayLabs/SmartRedis/pull/352
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -8,6 +8,7 @@ To be released at some future point in time
 
 Description
 
+- Fixed a memory leak in the C layer
 - Remove debug output in pybind layer
 - Update Hiredis version to 1.1.0
 - Enable parallel build for the SmartRedis examples
@@ -23,6 +24,7 @@ Description
 
 Detailed Notes
 
+- Fixed memory leaked from the C layer when calling get_string_option() (PR357_)
 - Remove debug output in pybind layer associated with put_dataset (PR352_)
 - Updated to the latest version of Hiredis (1.1.0) (PR351_)
 - Enable parallel build for the SmartRedis examples by moving utility Fortran code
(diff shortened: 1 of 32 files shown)


src/python/module/smartredis/dataset_utils.py
```diff
@@ -94,13 +106,13 @@ def add_metadata_for_xarray(
...
+        if isinstance(data_names, str):
             data_names = [data_names]
-        if type(dim_names) == str:
+        if isinstance(dim_names, str):
             dim_names = [dim_names]
-        if type(coord_names) == str:
+        if isinstance(coord_names, str):
             coord_names = [coord_names]
-        if type(attr_names) == str:
+        if isinstance(attr_names, str):
```
MattToast, line 115: Good catch!! TY for cleaning these up!!

src/python/module/smartredis/dataset_utils.py
```diff
@@ -131,7 +143,7 @@ def add_metadata_for_xarray(
 
     @staticmethod
     @_requires_xarray
-    def transform_to_xarray(dataset):
+    def transform_to_xarray(dataset: Dataset)->t.Dict:
```
MattToast, line 146: Am I right in assuming that this can be narrowed to
```py
def transform_to_xarray(dataset: Dataset) -> t.Dict[str, xr.DataArray]:
```
?

src/python/module/smartredis/dataset_utils.py
```diff
@@ -41,10 +49,9 @@
 
 # ----helper decorators -----
 
-
-def _requires_xarray(fn):
+def _requires_xarray(fn: t.Callable[_PR, _RT])->TFunc:
```
MattToast, line 52: Looks like the generic `_PR` and `_RT` are not be followed into the return val here bc of the type allias (similar problem to what we were seeing yesterday)
```py
@_requires_xarray
def dummy(a: int, b: str) -> float:
    ...
reveal_type(dummy)  # reveals `(*Any, **Any) -> Any`
``` 
could we replace this with
```py
def _requires_xarray(fn: t.Callable[_PR, _RT]) -> t.Callable[_PR, _RT]:
```
mypy seems to be happier with that (at least for my little demo fn above)

src/python/module/smartredis/logcontext.py
```diff
@@ -42,13 +41,13 @@ def __init__(self, context):
         self._name = context
 
     @property
-    def _logcontext(self):
+    def _logcontext(self)->t.Type[SRObject]:
```
MattToast, line 44: Does should the return type here be `PySRObject`? That appears to be the intended type of `self._srobject` (currently this is getting past the type check bc `PySRObject` aliases to `Any` bc we do not have stubs for `smartredis.smartredisPy`)

src/python/module/smartredis/util.py
```diff
@@ -64,7 +71,7 @@ def metadata_from_numpy(array):
         raise TypeError(f"Incompatible metadata type provided {dtype}")
 
     @staticmethod
-    def from_string(type_name):
+    def from_string(type_name: str)->type:
```
MattToast, line 74: There is very, very, very subtle difference (that I do not fully remember) between `type` and `t.Type` when used without `[]`. I'm also not fully confident that that this syntax is accepted by py3.7 or 3.8 (3.9 was when the `list[T] == t.List[T]` syntax was introduced, I imagine that's true for `type` as well).

For those reasons, just for safety, can we change the return type here to `t.Type`

src/python/module/smartredis/util.py
```diff
@@ -94,7 +103,7 @@ def init_default(default, init_value, expected_type=None):
         raise TypeError(f"Argument was of type {type(init_value)}, not {expected_type}")
     return init_value
 
-def exception_handler(func):
+def exception_handler(func: t.Callable[_PR, _RT])->TFunc:
```
MattToast, line 106: Similar "allias doesn't look like it is being followed" issue here. can we also make this
```py
def exception_handler(func: t.Callable[_PR, _RT]) -> t.Callable[_PR, _RT]:
```

MattToast (changes requested):
Looks great Just a couple of touch ups, and a request to run
```sh
$ black src/python/module/smartredis/
```
to conform to the pep8 preference of 
```py
def fn() -> T:
      # ^  ^ Spaces around return type
    ...
```

pyproject.toml
```diff
@@ -26,3 +26,30 @@ exclude = '''
...
+
+# Safety/Upgrading Mypy
+warn_unused_ignores = true
+# warn_redundant_casts = true # not a per-module setting?
+
+[[tool.mypy.overrides]]
+# Ignore packages that are not used or not typed
+module = [
+  "smartredis.smartredisPy",
+#  "smartsim._core.launcher.step.*", # hints incomplete
```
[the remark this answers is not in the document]
billschereriii: Good catch. I left that in as an example of formatting and meant to nuke it

doc/changelog.rst
```diff
@@ -52,13 +54,11 @@ users need not import `iso_c_binding` in their own applications (PR340_)
 - New pip-install target in Makefile will be a dependency of the lib target going forward so that users don't have to manually pip install SmartRedis in the future (PR330_)
 - Added ConfigOptions class and API, which will form the backbone of multiDB support (PR303_)
 
+.. _PR361: https://github.com/CrayLabs/SmartRedis/pull/361
 .. _PR360: https://github.com/CrayLabs/SmartRedis/pull/360
 .. _PR359: https://github.com/CrayLabs/SmartRedis/pull/359
-.. _PR358: https://github.com/CrayLabs/SmartRedis/pull/358
```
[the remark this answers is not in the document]
billschereriii: Probably a merge mishap. Thanks for the catch

src/python/module/smartredis/client.py
```diff
@@ -332,7 +338,7 @@ def set_function_multigpu(self, name, function, first_gpu, num_gpus):
         self._client.set_script_multigpu(name, fn_src, first_gpu, num_gpus)
 
     @exception_handler
-    def set_script(self, name, script, device="CPU"):
+    def set_script(self, name: str, script: str, device: str="CPU")->None:
```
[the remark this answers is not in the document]
billschereriii: We let users screw up the casing on these without penalty so it gets a bit hairy. I'll keep this in mind for the future though

src/python/module/smartredis/client.py
```diff
@@ -448,7 +458,13 @@ def get_script(self, name):
         return script
 
     @exception_handler
-    def run_script(self, name, fn_name, inputs, outputs):
+    def run_script(
+        self,
+        name: str,
+        fn_name: str,
+        inputs: t.List[str],
+        outputs: t.List[str]
```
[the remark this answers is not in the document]
billschereriii: Can totally do that. The typecheck() call would have to be updated to support this though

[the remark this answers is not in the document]
billschereriii: I'll open a ticket to consider this one

src/python/module/smartredis/dataset.py
```diff
@@ -164,29 +165,31 @@ def add_meta_string(self, name, data):
         self._data.add_meta_string(name, data)
 
     @exception_handler
-    def get_meta_scalars(self, name):
+    def get_meta_scalars(self, name: str)->t.Union[t.List[int], t.List[float]]:
```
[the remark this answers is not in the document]
billschereriii: Exactly as intended

src/python/module/smartredis/dataset_utils.py
billschereriii: Maybe. I think we'll be refining the type hints as we extend the code once Julia comes onboard

src/python/module/smartredis/dataset_utils.py
billschereriii: Done here and in util.py

src/python/module/smartredis/util.py
billschereriii: done

src/python/module/smartredis/logcontext.py
billschereriii: Good catch!

src/python/module/smartredis/client.py
```diff
@@ -575,16 +601,16 @@ def get_model(self, name):
...
-    ):
+        name: str,
+        model: bytes,
+        backend: str,
+        device: str="CPU",
+        batch_size: int=0,
+        min_batch_size: int=0,
+        tag: str="",
+        inputs: t.Optional[t.Union[str, t.List[str]]]=None,
+        outputs: t.Optional[t.Union[str, t.List[str]]]=None
```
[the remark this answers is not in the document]
billschereriii: Good catch! Done

MattToast (approved):
LGTM!!

PR #364, opened 2023-06-30
Reuse existing Redis connection when executing CLUSTER MAP command

billschereriii:
Title says it all


mellis13 (approved):
Thanks for putting in this improvement -- lgtm!

PR #371, opened 2023-07-06
mitigate lint errors, apply formatters

ankona:
- Mitigate linting/formatting errors that have crept in over time.
- Add execution of linter to github workflow
- Apply black formatter 
- Apply isort formatter


src/python/module/smartredis/error.py
```diff
@@ -61,6 +61,8 @@ def _check_error(cpp_error: str, method: str = "", key: str = "") -> str:
         msg = ""
         if method:
             msg = f"{method} execution failed\n"
+        if key:
+            msg += f"key: {key}\n"
```
billschereriii, line 65: This will usually be redundant with the underlying Redis error message, so I think this change should not happen

src/python/module/smartredis/util.py
```diff
@@ -24,16 +24,18 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from .error import *
+import typing as t
 from functools import wraps
+
+import numpy as np
+
```
billschereriii, line 31: I don't understand how we got to this ordering

src/python/module/smartredis/util.py
```diff
@@ -102,7 +104,8 @@ def init_default(
     if init_value is None:
         return default
     if expected_type is not None and not isinstance(init_value, expected_type):
-        raise TypeError(f"Argument was of type {type(init_value)}, not {expected_type}")
+        msg = f"Argument was of type {type(init_value)}, not {expected_type}"
+        raise TypeError(msg)
```
billschereriii, line 108: What was the point of this change?

src/python/module/smartredis/util.py
ankona: me neither. i didn't really inspect the result of isort but I'm glad to see if another preferred ordering doesn't trigger a change

src/python/module/smartredis/error.py
ankona: the linter complains about the unused variable. can i get rid of the unused method param, too?

src/python/module/smartredis/dataset_utils.py
```diff
@@ -221,3 +217,23 @@ def transform_to_xarray(dataset: Dataset) -> t.Dict:
...
+        :return: a dictionary of keys as the data field name and the
+        value as the built Xarray DataArray constructed using
+        fieldnames and appropriately formatted metadata.
+        rtype: dict
+        """
+        typecheck(dataset, "dataset", Dataset)
+
+        coord_final = DatasetConverter._find_coord_vars(dataset)
+        ret_xarray = DatasetConverter._construct_xarray(dataset, coord_final)
+        return ret_xarray
```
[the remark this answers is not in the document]
ankona: linter complained about too many local variables. i split the function into two to reduce the # of locals in a given space

tests/python/test_logging.py
```diff
@@ -24,41 +24,51 @@
...
-])
+
+@pytest.mark.parametrize("log_level", [LLQuiet, LLInfo, LLDebug, LLDeveloper])
 def test_logging_client(use_cluster, context, log_level):
     c = Client(None, use_cluster, logger_name=context)
     c.log_data(log_level, f"This is data logged from a client ({log_level.name})")
-    c.log_warning(log_level, f"This is a warning logged from a client ({log_level.name})")
+    c.log_warning(
+        log_level, f"This is a warning logged from a client ({log_level.name})"
+    )
```
billschereriii, line 51: client and string are the same length.  I don't see why all the string ones above got split to multiple lines, but only some of the client ones did. Can we just keep all these on single lines?

src/python/module/smartredis/error.py
billschereriii: Let's check to see if anyone ever passes the key to the constructor first. If not (which I expect), then definitely

billschereriii: This is the only change that I'm really concerned about btw

billschereriii (changes requested):
Generally looks very good. I'd like to get the key parameter for RedisReplyError resolved before approving

src/python/module/smartredis/util.py
ankona: Busted out [isort docs](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html) and it does make sense, now. 

Default sort ordering is: `FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER`

numpy is moved out of the stdlib section into a third-party grouping, per that default.

tests/python/test_logging.py
ankona: I added some column indicators at 80, 88, and 94 characters to illustrate:

image

The `log_warning` calls are definitely over the 88 character limit. I can manually reformat this file for now. Without further tweaks, it would get picked up again on the next `make format`:

image

For now, I reformatted & disabled re-formatting in this file with `# fmt: off`

src/python/module/smartredis/error.py
ankona: Searched for any usage of the parameter. Found none in smartredis or smartsim. Removed parameter.

PR #376, opened 2023-07-12
Update language support chart

billschereriii:
Update language support chart to match updates that happened in the last release


MattToast (approved):
Good catch! LGTM!!

PR #377, opened 2023-07-13
[NO_MERGE] Debugging code for put/get tensor methods in Python and C++

billschereriii:
WARNING, there is currently a break in the develop branch of SmartRedis that will cause hard failures in the event that an exception is transmitted from the C++ layer to the Python layer. While this will be resolved shortly, I wanted to go ahead and put this test code up anyway as it might be useful despite this caveat.


mellis13:
The issue has been resolved so this draft PR is no longer needed for debugging.

PR #378, opened 2023-07-13
Fix bad kwarg when launching server via UDS

ankona:
When launching a server with UDS enabled, a keyword arg typo causes an error to occur.

The current code contains:

````
def prepare_uds_socket(udsport):
    """Sets up the UDS socket"""
    if udsport is None:
        return # Silently bail
    uds_abs = pathlib.Path(udsport).resolve()
    basedir = uds_abs.parent
    basedir.mkdir(exists_okay=True)
    ...
````

The defect is in the line `basedir.mkdir(exists_okay=True)` - the keyword should be `exist_ok`. See [pathlib docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir)

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -15,10 +15,12 @@ Description
 Detailed Notes
 
 - Update language support matrix in documentation to reflect updates from the last release (PR379_)
+- Fix typo causing startup failure in utility script for unit tests (PR378_)
 - Deleted obsolete build and testing files that are no longer needed with the new build and test system (PR366_)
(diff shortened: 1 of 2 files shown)


billschereriii (approved):
Please add a changelog update before merging

billschereriii (changes requested):
Given that UDS isn't supported on MacOS, the new tests will cause failures on that platform.

I would prefer that if any tests are added, they be done in a new PR

ashao (approved):
Thanks for the fix!

billschereriii (approved):
LGTM!

PR #381, opened 2023-07-13
Make testing more robust

billschereriii:
Make sure to shut down Redis server even if tests fail. 
Do so while still reporting the result of testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -367,7 +367,7 @@ define run_smartredis_tests_with_uds_server
 		$(SKIP_DOCKER) $(SKIP_PYTHON) $(SKIP_FORTRAN) \
 		--build $(SR_BUILD) --sr_fortran $(SR_FORTRAN) $(1) ; \
 	(testresult=$$?; \
-	echo "Shutting down standalone Redis server with Unix Domain Socket support" &&
+	echo "Shutting down standalone Redis server with Unix Domain Socket support" && \
 	python utils/launch_redis.py --port $(SR_TEST_PORT) --nodes 1 \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -320,9 +320,11 @@ define run_smartredis_tests_with_standalone_server
 	echo "Running standalone tests" && \
 	PYTHONFAULTHANDLER=1 python -m pytest $(SR_TEST_PYTEST_FLAGS) $(COV_FLAGS) \
 		$(SKIP_DOCKER) $(SKIP_PYTHON) $(SKIP_FORTRAN) \
-		--build $(SR_BUILD) --sr_fortran $(SR_FORTRAN) $(1) && \
+		--build $(SR_BUILD) --sr_fortran $(SR_FORTRAN) $(1)  ; \
+	(testresult=$$?; \
 	echo "Shutting down standalone Redis server" && \
 	python utils/launch_redis.py --port $(SR_TEST_PORT) --nodes 1 --stop && \
+	exit $$testresult) && \
```
ashao, line 327: I think this would be slightly more bash-like:
```
test $$testresult -eq 0 || echo "Tests failed"; exit $$test_result
```

ashao (changes requested):
Small tweak, but I want to make sure that you agree from a readability/correctness standpoint

Makefile
billschereriii: I like that, will adopt

ashao (approved):
Looks great thanks Bill!

MattToast (approved):
LGTM!!

PR #382, opened 2023-07-14
mitigate lint errors, apply formatters

ankona:
- Mitigate linting/formatting errors that have crept in over time.
- Add execution of linter to github workflow
- Apply black formatter 
- Apply isort formatter


doc/changelog.rst
```diff
@@ -16,11 +16,13 @@ Detailed Notes
 
 - Update language support matrix in documentation to reflect updates from the last release (PR379_)
 - Fix typo causing startup failure in utility script for unit tests (PR378_)
+- Update pylint configuration and mitigate issues, run in CI pipeline (PR371_)
```
billschereriii, line 19: Add a brief change description at the top as well

setup.cfg
```diff
@@ -43,7 +43,7 @@ dev =
     pytest-cov==2.10.1
     black==20.8b1
     isort==5.6.4
-    pylint==2.6.0
+    pylint>=2.10.0
```
billschereriii, line 46: Current version is 2.17. We should use that

billschereriii (changes requested):
Looking good. A few minor changes

setup.cfg
ankona: 2.17 actually causes a dependency issue with something else, iirc. I went back to a min version that worked and didn't show that collision

billschereriii: If you could open a ticket about this, it would be good to track. No further update needed here then

ankona: re-tested & saw no issue(s). updating.

billschereriii (approved):
LGTM, thanks!

PR #386, opened 2023-07-25
Static branch fixups

billschereriii:
Assorted changes in support of static builds:

- added pytext fixture for link mode
- static and shared builds are now done in separate trees since CMake doesn't seem to consider them as distinct configurations
- Test files updated to build and run with the new paths
- Examples update to build with the new path and run with CWD=source file location directory
- CI/CD updated to perform static builds of examples on non-MacOS platforms


Makefile
```diff
@@ -152,65 +152,70 @@ test-deps-gpu: test-deps
 
 # help: build-tests                    - build all tests (C, C++, Fortran)
 .PHONY: build-tests
+build-tests: test-deps
```
ashao, line 155: I think I'm actually walking back this suggestion that I had made. I think building the tests are useful enough (without running them) that maybe we just add the dependency of the CPP tests on `catch` to this list?

examples/parallel/fortran/CMakeLists.txt
```diff
@@ -42,33 +43,75 @@ if(NOT DEFINED SMARTREDIS_INSTALL_PATH)
 endif()
 
 # Locate dependencies
+#  . MPI
 find_package(MPI REQUIRED)
 IF(NOT MPI_Fortran_FOUND)
 	message(FATAL_ERROR "Could not find Fortran MPI components")
 endif()
+#  . Main SmartRedis Library (C++ based)
```
ashao, line 51: More accurate to say C/C++

examples/parallel/fortran/CMakeLists.txt
```diff
@@ -42,33 +43,75 @@ if(NOT DEFINED SMARTREDIS_INSTALL_PATH)
...
+add_library(smartredis-main ${SMARTREDIS_LINK_MODE} IMPORTED)
 find_library(SR_LIB ${SMARTREDIS_LIB}
     PATHS ${SMARTREDIS_INSTALL_PATH}/lib NO_DEFAULT_PATH
     REQUIRED
     ${SMARTREDIS_LINK_MODE}
 )
+set_target_properties(smartredis-main PROPERTIES
+    IMPORTED_LOCATION ${SR_LIB}
+)
+#  . SmartRedis Fortran Library (C++ based)
```
ashao, line 61: `(C++ based)` is not innacurate

.github/workflows/run_tests.yml
```diff
@@ -145,6 +145,18 @@ jobs:
                 TF /usr/lib/redis/modules/backends/redisai_tensorflow/redisai_tensorflow.so \
                 TORCH /usr/lib/redis/modules/backends/redisai_torch/redisai_torch.so" >> $GITHUB_ENV
 
+      # Check the static build by compiling and linking examples
```
ashao, line 148: While I know that it's "free" to run our tests on Github Actions, what do you think about doing a static build first, then doing a make clean, and then continuing on?

I don't foresee a problem with actually executing the tests once they link correctly.

ashao (changes requested):
Thanks for tracking down all the CMake magic to make this happen. Some very small comments (except potentially the one about test changes).

.github/workflows/run_tests.yml
ashao: Disregard. I see this is what you're already doing

examples/parallel/fortran/CMakeLists.txt
billschereriii: Fixed

examples/parallel/fortran/CMakeLists.txt
billschereriii: Fixed

Makefile
billschereriii: Skipping per discussion

.github/workflows/run_tests.yml
billschereriii: static and shared build in separate directories, so the risk of cross-contagion is minimal -- and if it happens, we want to know about it anyway

ashao (approved):

PR #390, opened 2023-08-09
Extended section on dataset aggregation

billschereriii:
Expanded the documentation section on DataSet aggregation and moved it to a new Advanced Topics page

### DIFF
--- a/doc/advanced_data_structures.rst
+++ b/doc/advanced_data_structures.rst
@@ -1,0 +1,117 @@
+******************************
+Advanced Data Structure Topics
+******************************
+
+This page of documentation is reserved for advanced topics in SmartRedis
+data structures that may not be needed for all users.
+
+.. _advanced_data_structures_dataset_aggregation:
+
+Dataset Aggregation
+===================
+
+In addition to the ability to work with individual datasets, SmartRedis offers
+users the manage lists of datasets and work with them collectively. This is
+particularly useful for producer/consumer relationships.
+
+The DataSet Aggregation API stores multiple ``DataSet`` objects
+on one or more database nodes through an interface referred to as
+``aggregation lists``.
+An ``aggregation list`` in SmartRedis stores references to
+``DataSet`` objects that are stored in the database.  The SmartRedis client
+in one application can append ``DataSet`` objects to the ``aggregation list``;
+subsequently, ``SmartRedis`` clients in the same application or a different
+application can retrieve some or all of the ``DataSet`` objects referenced
+in that ``aggregation list``.
+
+Appending to a DataSet
+----------------------
+
+For example, the C++ client function to append a ``DataSet`` to an
+aggregation list is shown below:
+
+.. code-block:: cpp
+
+    # C++ aggregation list append interface
+    void append_to_list(const std::string& list_name,
+                        const DataSet& dataset);
+
+The above function will append the provided ``DataSet`` to the
+``aggregation list``, which can be referenced in all user-facing functions
+by the provided list name.  Note that a list can be appended by
+any client in the same or different application.  Implicitly, all
+DataSets, when appended, are added to the end of the list. If the list does not
+already exist, it is automatically created.
+
+Retrieving DataSets from an aggregation list
+--------------------------------------------
+
+To retrieve ``aggregation list`` contents,
+the SmartRedis ``Client`` method provides an API function that
+will return an iterable container with all of the ``DataSet`` objects
+appended to the ``aggregation list``.  For example, the C++ client
+function to retrieve the entire ``aggregation list`` contents is shown below:
+
+.. code-block:: cpp
+
+    # C++ aggregation list retrieval interface
+    std::vector<DataSet> get_datasets_from_list(const std::string& list_name);
+
+It is also possible to retrieve a subset of the DataSets within an aggregation
+list:
+
+.. code-block:: cpp
+
+    # C++ aggregation list subset retrieval interface
+    std::vector<DataSet> get_dataset_list_range(const std::string& list_name,
+                                                    const int start_index,
+                                                    const int end_index);
+
+An application can determine how many DataSets are in an aggregation List
+via the following API:
+
+.. code-block:: cpp
+
+    # C++ aggregation list length query
+    int get_list_length(const std::string& list_name);
+
+Synchronization
+---------------
+
+A SmartRedis Client can use the following APIs to block until a
+predicate is matched on the length of the list:
+
+.. code-block:: cpp
+
+    # Block until the list reaches a specific length
+    bool poll_list_length(const std::string& name, int list_length,
+                            int poll_frequency_ms, int num_tries);
+
+    # Block until the list reaches or exceeds a specific length
+    bool poll_list_length_gte(const std::string& name, int list_length,
+                                int poll_frequency_ms, int num_tries);
+
+    # Block until the list is no longer than a specific length
+    bool poll_list_length_lte(const std::string& name, int list_length,
+                                int poll_frequency_ms, int num_tries);
+
+Other operations
+----------------
+
+Finally, aggregation lists may be copied, renamed, or deleted. Note
+that there is no synchronization support for these operations; performing
+these operations when multiple applications are accessing the list may
+lead to race conditions:
+
+.. code-block:: cpp
+
+    # Copy an aggregation list
+    void copy_list(const std::string& src_name,
+                    const std::string& dest_name);
+
+    # Rename an aggregation list
+    void rename_list(const std::string& src_name,
+                        const std::string& dest_name);
+
+    # Delete an aggregation list
+    void delete_list(const std::string& list_name);
--- a/doc/advanced_topics.rst
+++ b/doc/advanced_topics.rst
@@ -2,16 +2,16 @@
 Advanced Topics
 ***************
 
-This page of documentation is reserved for advanced topics in SmartRedis
-data structures that may not be needed for all users.
+This page of documentation is reserved for advanced topics
+that may not be needed for all users.
 
 .. _advanced_topics_dataset_aggregation:
 
 Dataset Aggregation
 ===================
 
-In addition to the ability to work with individual datasets, SmartRedis offers
-users the manage lists of datasets and work with them collectively. This is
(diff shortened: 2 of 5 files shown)


doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+An ``aggregation list`` in SmartRedis stores references to
+``DataSet`` objects that are stored in the database.  The SmartRedis client
+in one application can append ``DataSet`` objects to the ``aggregation list``;
+subsequently, ``SmartRedis`` clients in the same application or a different
+application can retrieve some or all of the ``DataSet`` objects referenced
+in that ``aggregation list``.
+
+Appending to a DataSet
+----------------------
+
```
ashao, line 29: We need to highlight, emphasize, and underscore that users need to put the dataset into the database _before_ they can append it to the list.

ashao: In the future, I wonder if we should do the `put_dataset` for users with an optional `overwrite` flag if the dataset already exists.

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+``aggregation list``, which can be referenced in all user-facing functions
+by the provided list name.  Note that a list can be appended by
+any client in the same or different application.  Implicitly, all
+DataSets, when appended, are added to the end of the list. If the list does not
+already exist, it is automatically created.
+
+Retrieving DataSets from an aggregation list
+--------------------------------------------
+
+To retrieve ``aggregation list`` contents,
```
ashao, line 49: To retrieve the contents of an ``aggregation list``

ashao (approved):
Some very minor comments, but otherwise thank you for expanding this section!

doc/advanced_data_structures.rst
billschereriii: Good point, I'll make sure to capture that

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
+******************************
+Advanced Data Structure Topics
```
mellis13, line 2: I think part of the feedback on the original documentation is that dataset aggregation was really buried in the docs.  I wonder if we should move it to the top level under ``Dataset`` in the left nav bar and do ``Dataset aggregation``

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+.. _advanced_data_structures_dataset_aggregation:
+
+Dataset Aggregation
+===================
+
+In addition to the ability to work with individual datasets, SmartRedis offers
+users the manage lists of datasets and work with them collectively. This is
+particularly useful for producer/consumer relationships.
+
+The DataSet Aggregation API stores multiple ``DataSet`` objects
```
mellis13, line 17: Maybe change to ``The DataSet Aggregation API manages references to multiple ``DataSet`` objects``

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+on one or more database nodes through an interface referred to as
+``aggregation lists``.
+An ``aggregation list`` in SmartRedis stores references to
+``DataSet`` objects that are stored in the database.  The SmartRedis client
+in one application can append ``DataSet`` objects to the ``aggregation list``;
+subsequently, ``SmartRedis`` clients in the same application or a different
+application can retrieve some or all of the ``DataSet`` objects referenced
+in that ``aggregation list``.
+
+Appending to a DataSet
```
mellis13, line 27: This seems a little off.  Maybe more accurate to say "Appending a DataSet" or "Appending to a DataSet aggregation list"

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+For example, the C++ client function to append a ``DataSet`` to an
+aggregation list is shown below:
+
+.. code-block:: cpp
+
+    # C++ aggregation list append interface
+    void append_to_list(const std::string& list_name,
+                        const DataSet& dataset);
+
+The above function will append the provided ``DataSet`` to the
```
mellis13, line 39: Maybe change to "The above function will append a reference to the ..."

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+``aggregation list``, which can be referenced in all user-facing functions
+by the provided list name.  Note that a list can be appended by
+any client in the same or different application.  Implicitly, all
+DataSets, when appended, are added to the end of the list. If the list does not
+already exist, it is automatically created.
+
+Retrieving DataSets from an aggregation list
```
mellis13, line 46: Maybe "via" is more accurate than "from" here?

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+--------------------------------------------
+
+To retrieve ``aggregation list`` contents,
+the SmartRedis ``Client`` method provides an API function that
```
mellis13, line 50: Is "method" needed here?

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+
+It is also possible to retrieve a subset of the DataSets within an aggregation
+list:
+
+.. code-block:: cpp
+
+    # C++ aggregation list subset retrieval interface
+    std::vector<DataSet> get_dataset_list_range(const std::string& list_name,
+                                                    const int start_index,
+                                                    const int end_index);
```
mellis13, line 68: Do we support the negative number indexing?  If so, do we want to point that out?

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+    bool poll_list_length_lte(const std::string& name, int list_length,
+                                int poll_frequency_ms, int num_tries);
+
+Other operations
+----------------
+
+Finally, aggregation lists may be copied, renamed, or deleted. Note
+that there is no synchronization support for these operations; performing
+these operations when multiple applications are accessing the list may
+lead to race conditions:
```
mellis13, line 104: This is a really great point to bring up!  I wonder if we should further clarify that some of the operations are compound operations (and which ones).

doc/advanced_data_structures.rst
```diff
@@ -0,0 +1,117 @@
...
+    # Copy an aggregation list
+    void copy_list(const std::string& src_name,
+                    const std::string& dest_name);
+
+    # Rename an aggregation list
+    void rename_list(const std::string& src_name,
+                        const std::string& dest_name);
+
+    # Delete an aggregation list
+    void delete_list(const std::string& list_name);
```
mellis13, line 117: (general comment, not a comment on this line)  I wonder if we should include a mention about the handling of aggregation lists within and across ensembles.

mellis13 (changes requested):
Thanks for improving the docs!  Just a couple of small questions/comments.

doc/advanced_data_structures.rst
billschereriii: More precise, I like it. Done

doc/advanced_data_structures.rst
billschereriii: Good catch, fixed

doc/advanced_data_structures.rst
billschereriii: Opened a ticket for the suggestion

doc/advanced_data_structures.rst
billschereriii: Fixed

doc/advanced_data_structures.rst
billschereriii: Renamed section to "Advanced topics" to match original intent. This will result in these topics being readily accessible as top-level items on the right and thus easy to find.

doc/advanced_data_structures.rst
billschereriii: Added discussion of negative indexing

doc/advanced_data_structures.rst
billschereriii: I don't think these operations being compound (non-atomic) really affects the race conditions described elsewhere

doc/advanced_data_structures.rst
billschereriii: Will open a separate ticket to make sure that any interactions between dataset aggregation and ensembles are covered. I think the only interaction should be with prefixing, but this needs to be researched.

ashao (approved):
I agree with the changes you made in response to Matt's comments. Thanks again!

doc/advanced_topics.rst
```diff
@@ -0,0 +1,126 @@
+***************
+Advanced Topics
+***************
+
+This page of documentation is reserved for advanced topics in SmartRedis
+data structures that may not be needed for all users.
```
mellis13, line 6: Is "data structures" here needed?

doc/advanced_topics.rst
```diff
@@ -0,0 +1,126 @@
...
+
+.. _advanced_topics_dataset_aggregation:
+
+Dataset Aggregation
+===================
+
+In addition to the ability to work with individual datasets, SmartRedis offers
```
mellis13, line 13: This sentence seems like it got messed up a bit in the revision.  Maybe should be "In addition to the ability to work with individual datasets, SmartRedis enables managing lists of datasets and working with them collectively."

doc/advanced_topics.rst
```diff
@@ -0,0 +1,126 @@
...
+    std::vector<DataSet> get_dataset_list_range(const std::string& list_name,
+                                                    const int start_index,
+                                                    const int end_index);
+
+The start_index and end_index may be specified as negative numbers.
+In this case, the offset is from the most recently appended DataSets.
+For example, an offset of -1 is the last element in the list, and -3 is the
+antepenultimate DataSet.
+
+An application can determine how many DataSets are in an aggregation List
```
mellis13, line 78: List -> list

mellis13 (approved):

doc/advanced_topics.rst
billschereriii: Good catch, thanks!

doc/advanced_topics.rst
billschereriii: Another good catch, thanks

PR #396, opened 2023-08-30
lcov update

billschereriii:


.github/workflows/run_tests.yml
```diff
@@ -108,6 +108,25 @@ jobs:
           echo "CXX=icpx" >> $GITHUB_ENV &&
           echo "FC=ifort" >> $GITHUB_ENV
 
+      # Set up perl environment for LCOV
+      - uses: actions/checkout@v3
+      - name: Setup perl
+        uses: shogo82148/actions-setup-perl@v1
+        with:
+          perl-version: '5.30'
+          install-modules: Memory::Process
```
ashao, line 117: Any particular reason why this couldn't be installed with apt?

Makefile
```diff
@@ -596,12 +596,12 @@ third-party/catch/single_include/catch2/catch.hpp:
...
-third-party/lcov/install/usr/local/bin/lcov:
+lcov: third-party/lcov/install/bin/lcov
+third-party/lcov/install/bin/lcov:
 	@mkdir -p third-party
 	@cd third-party && \
 	git clone $(LCOV_URL) lcov --branch $(LCOV_VER) --depth=1
 	@cd third-party/lcov && \
 	mkdir -p install && \
-	CC=gcc CXX=g++ DESTDIR="install/" make install && \
+	CC=gcc CXX=g++ make PREFIX=`pwd`/install/ install && \
```
ashao, line 606: I'd slightly prefer `$PWD` or even `$(pwd)`

ashao (approved):
Very minor comments. Otherwise it looks good to me!

.github/workflows/run_tests.yml
billschereriii: library not found, so we had to do it this way

Makefile
billschereriii: No objections to that. I just went with what I knew :)

billschereriii: Oddly, $(PWD) and $(CURDIR) are both failing

PR #402, opened 2023-09-20
Update RedisAI version to v1.2.7

billschereriii:
Unit tests were testing against v1.2.3
CI/CD was testing against versions 1.2.4 and 1.2.5

This update enables model chunking support


MattToast (approved):
LGTM!!

PR #404, opened 2023-09-26
Model chunking support

billschereriii:


include/client.h
```diff
@@ -1269,6 +1269,19 @@ class Client : public SRObject
...
 
+        /*!
+        *   \brief Reconfigure the chunking size that Redis uses for model
+        *          serialization, replication, and the model_get command.
+        *   \details This method triggers the AI.CONFIG method in the Redis
+        *             database to change the model chunking size. The default
+        *             size of 511MB should be fine for most applications, so
+        *             it is expected to be very rare that a client calls this
+        *             method.
+        *   \param chunk_size The new chunk size
```
ashao, line 1280: I'm pretty sure the unit is bytes, but let's call that out specifically in the docstring

include/pyclient.h
```diff
@@ -925,6 +925,19 @@ class PyClient : public PySRObject
...
 
+        /*!
+        *   \brief Reconfigure the chunking size that Redis uses for model
+        *          serialization, replication, and the model_get command.
+        *   \details This method triggers the AI.CONFIG method in the Redis
+        *             database to change the model chunking size. The default
+        *             size of 511MB should be fine for most applications, so
+        *             it is expected to be very rare that a client calls this
+        *             method.
+        *   \param chunk_size The new chunk size
```
ashao, line 936: Same here

include/redisserver.h
```diff
@@ -520,6 +520,33 @@ class RedisServer {
...
+
+        /*!
+        *   \brief Reconfigure the chunking size that Redis uses for model
+        *          serialization, replication, and the model_get command.
+        *   \details This method triggers the AI.CONFIG method in the Redis
+        *             database to change the model chunking size. The default
+        *             size of 511MB should be fine for most applications, so
+        *             it is expected to be very rare that a client calls this
+        *             method.
+        *   \param chunk_size The new chunk size
```
ashao, line 537: Units here

ashao (approved):
Minor comment about adding in the units. Also as we agree offline, in the documentation please mention that setting the chunksize is generally not needed, even by advanced users; instead highlight that now we support users uploading models >=512MB.

include/client.h
billschereriii: Addressed in all 6 places where needed

include/pyclient.h
billschereriii: Addressed in all 6 places where needed

include/redisserver.h
billschereriii: Addressed in all 6 places where needed

PR #406, opened 2023-10-02
Expose MINBATCHTIMEOUT parameter in set_model interface

billschereriii:


ashao (changes requested):
Thanks for turning that around so quick! Will definitely lead to easy performance gains.

Only one request from me
- Throw a warning if `MINBATCHSIZE>0` and `MINBATCHTIMEOUT==0` to nudge the user. Something along the lines of `min_batch_timeout was set with a non-zero min_batch_size. Performance may improve by passing in a small value (~10ms) for min_batch_timeout`
- Throw an exception if `MINBATCHTIMEOUT>0 and (MINBATCHSIZE==0 or BATCHSIZE==0)` with the message that `min_batch_size and batch_size must be non-zero if min_batch_timeout>0`

Add those tests to ensure we're throwing those to `test_errors.py`.

.github/workflows/run_tests.yml
```diff
@@ -67,6 +67,13 @@ jobs:
       GCC_V: ${{ matrix.compiler }} # used when the compiler is gcc/gfortran
 
     steps:
+      # Free up some disk space
+      run: |
```
ashao: Ooof really these are the kind of games that we have to play?Q

ashao (approved):
Really appreciate the new usages exceptions/warnings. Also nice work on adding the test to capture the warning

PR #408, opened 2023-10-04
Update RedisAI version in post commit checking

billschereriii:
Need to update post-commit checking RedisAI version to v1.2.7 to support fetch of model chunking size parameter


ashao (approved):
Looks great! Thanks.

PR #412, opened 2023-10-12
Run examples on commit

billschereriii:
Fix bug with running examples introduced in multiDB PR.
Move test of static builds and examples to on-commit, not post-merge


ashao (approved):
Thanks for fixing up those tests!

PR #413, opened 2023-10-13
Use Redis++ ConnectionObject to set TCP socket timeout

billschereriii:
TCP default timeout is ridiculously long, so if we attempt to connect to a missing database, we're stuck waiting on the TCP timeout repeatedly. This will improve our reponsiveness in such cases

### DIFF
--- a/src/cpp/redis.cpp
+++ b/src/cpp/redis.cpp
@@ -728,16 +728,15 @@ inline void Redis::_connect(SRAddress& db_address)
     sw::redis::ConnectionOptions connectOpts;
     if (db_address._is_tcp) {
         connectOpts.host = db_address._tcp_host;
-        std::cout << "Connecting to host: " + connectOpts.host + "; ";
         connectOpts.port = db_address._tcp_port;
-        std::cout << "Connecting to port: " + std::to_string(connectOpts.port) + "\n";
         connectOpts.type = sw::redis::ConnectionType::TCP;
     }
     else {
         connectOpts.path = db_address._uds_file;
         connectOpts.type = sw::redis::ConnectionType::UNIX;
     }
-    connectOpts.socket_timeout = std::chrono::milliseconds(250);
+    connectOpts.socket_timeout = std::chrono::milliseconds(
+        _DEFAULT_SOCKET_TIMEOUT);
 
     // Connect
     for (int i = 1; i <= _connection_attempts; i++) {
--- a/src/cpp/redis.cpp
+++ b/src/cpp/redis.cpp
@@ -722,10 +722,22 @@ inline void Redis::_add_to_address_map(SRAddress& db_address)
 
 inline void Redis::_connect(SRAddress& db_address)
 {
+    // Build a connections object for this connection
+    // No need to repeat the build on each connection attempt
+    // so we do it outside the loop
+    sw::redis::ConnectionOptions connectOpts;
+    connectOpts.host = db_address._tcp_host;
+    if (db_address._is_tcp)
+        connectOpts.port = db_address._tcp_port;
+    else
+        connectOpts.path = db_address._uds_file;
+    connectOpts.socket_timeout = std::chrono::milliseconds(100);
+
+    // Connect
     for (int i = 1; i <= _connection_attempts; i++) {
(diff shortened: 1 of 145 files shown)


billschereriii:
Hmm.  A Redis connection based on the ConnectionOptions object doesn't seem to be stable. Will investigate further next week

src/cpp/redis.cpp
```diff
@@ -722,10 +722,28 @@ inline void Redis::_add_to_address_map(SRAddress& db_address)
 
 inline void Redis::_connect(SRAddress& db_address)
 {
+    // Build a connections object for this connection
+    // No need to repeat the build on each connection attempt
+    // so we do it outside the loop
+    sw::redis::ConnectionOptions connectOpts;
+    if (db_address._is_tcp) {
+        connectOpts.host = db_address._tcp_host;
+        std::cout << "Connecting to host: " + connectOpts.host + "; ";
```
ashao, line 731: Can we move this into the logger somehow instead of outputting directly to STDOUT

src/cpp/redis.cpp
```diff
@@ -722,10 +722,28 @@ inline void Redis::_add_to_address_map(SRAddress& db_address)
...
+        connectOpts.port = db_address._tcp_port;
+        std::cout << "Connecting to port: " + std::to_string(connectOpts.port) + "\n";
```
ashao, line 733: Same

src/cpp/rediscluster.cpp
```diff
@@ -1075,6 +1075,24 @@ inline CommandReply RedisCluster::_run(const Command& cmd, std::string db_prefix
 // Connect to the cluster at the address and port
 inline void RedisCluster::_connect(SRAddress& db_address)
 {
+    // Build a connections object for this connection
+    // No need to repeat the build on each connection attempt
+    // so we do it outside the loop
+    sw::redis::ConnectionOptions connectOpts;
+    if (db_address._is_tcp) {
+        connectOpts.host = db_address._tcp_host;
+        std::cout << "Connecting to host: " + connectOpts.host + "; ";
```
ashao, line 1084: Also here capture to logger, or eliminate the line

src/cpp/redis.cpp
```diff
@@ -722,10 +722,28 @@ inline void Redis::_add_to_address_map(SRAddress& db_address)
...
+        connectOpts.type = sw::redis::ConnectionType::TCP;
+    }
+    else {
+        connectOpts.path = db_address._uds_file;
+        connectOpts.type = sw::redis::ConnectionType::UNIX;
+    }
+    connectOpts.socket_timeout = std::chrono::milliseconds(250);
```
ashao, line 740: Let's make the 250 a #define somewhere

src/cpp/rediscluster.cpp
```diff
@@ -1075,6 +1075,24 @@ inline CommandReply RedisCluster::_run(const Command& cmd, std::string db_prefix
...
+        connectOpts.port = db_address._tcp_port;
+        std::cout << "Connecting to port: " + std::to_string(connectOpts.port) + "\n";
+        connectOpts.type = sw::redis::ConnectionType::TCP;
+    }
+    else {
+        throw SRInternalException(
+            "RedisCluster encountered a UDS request in _connect()");
+    }
+    connectOpts.socket_timeout = std::chrono::milliseconds(250);
```
ashao, line 1093: Same suggestion to move htis to a #define

ashao (changes requested):

src/cpp/redis.cpp
billschereriii: Debug text -- will remove

src/cpp/redis.cpp
billschereriii: same

src/cpp/rediscluster.cpp
billschereriii: same

src/cpp/redis.cpp
billschereriii: done -- good thought

ashao (approved):

PR #415, opened 2023-10-16
Update allowed special characters in SSDB name

ashao:
SmartSim allows users to declare their own name for a UDS socket file, however the `_check_ssdb_string` function is overly strict and throws an exception for two commonly used special characters. The modification here allows users to name their UDS socket file with `_` and `-`.


billschereriii (changes requested):
PR looks good, but please update the changelog in doc/changelog.rst

billschereriii (approved):
LGTM, Thanks for plugging this hole!

PR #419, opened 2023-10-24
Python Client constructor update

billschereriii:
Clean up Python Client constructor to detect invalid kwargs and remove deprecation warning for construction method that will be used primarily for driver scripts. Rename internal construction methods to reflect new expected usage patterns. Update test cases to remove invalid construction requests. Update docs.

### DIFF
--- a/src/python/module/smartredis/client.py
+++ b/src/python/module/smartredis/client.py
@@ -90,7 +90,12 @@ class Client(SRObject):
                 pyclient = self.__standard_construction(*a, **kw)
         super().__init__(pyclient)
 
-    def __address_construction(self, cluster, address=None, logger_name="Default"):
+    def __address_construction(
+        self,
+        cluster: bool,
+        address: t.Optional[str] = None,
+        logger_name: str = "Default"
+    ) -> PyClient:
         """Initialize a SmartRedis client
 
         This construction method is primarily intended for use by driver
@@ -121,7 +126,11 @@ class Client(SRObject):
         except (PybindRedisReplyError, RuntimeError) as e:
             raise RedisConnectionError(str(e)) from None
 
-    def __standard_construction(self, config_options=None, logger_name="Default"):  # pylint: disable=no-self-use
+    @staticmethod
+    def __standard_construction(
+        config_options: t.Optional[ConfigOptions] = None,
+        logger_name: str = "Default"
+    ) -> PyClient:
         """Initialize a RedisAI client
 
         The address of the Redis database is expected to be found in the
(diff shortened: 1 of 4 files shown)


src/python/module/smartredis/client.py
```diff
@@ -121,7 +121,7 @@ def __deprecated_construction(self, cluster, address=None, logger_name="Default"
         except (PybindRedisReplyError, RuntimeError) as e:
             raise RedisConnectionError(str(e)) from None
 
-    def __new_construction(self, config_options=None, logger_name="Default"):  # pylint: disable=no-self-use
+    def __standard_construction(self, config_options=None, logger_name="Default"):  # pylint: disable=no-self-use
```
MattToast, line 124: Not directly related this PR, but thought I should bring it up while we are here:

You should be able to remove the `# pylint: disable=...` comment by simply decorating this method with `@staticmethod` and removing the `self` parameter

src/python/module/smartredis/client.py
```diff
@@ -43,50 +42,60 @@
...
+            # Only kwargs in the call
+            if "address" in kw:
+                pyclient = self.__address_construction(*a, **kw)
+            else:
+                pyclient = self.__standard_construction(*a, **kw)
         super().__init__(pyclient)
 
-    def __deprecated_construction(self, cluster, address=None, logger_name="Default"):
-        """Initialize a RedisAI client (Deprecated)
+    def __address_construction(self, cluster, address=None, logger_name="Default"):
```
MattToast, line 93: Not desperately important, but if we could retroactively type hint this method that might be nice for future developers
```py
    def __address_construction(
        self,
        cluster: bool,
        address: t.Optional[str] = None,
        logger_name: str = "Default"
    ) -> PyClient:
```

src/python/module/smartredis/client.py
MattToast, line 124: Again, not desperately important, but if we could we retroactively type hint this while we are here?
```py
    def __standard_construction(
        self,
        config_options t.Optional[ConfigOptions] = None,
        logger_name: str = "Default"
    ) -> PyClient:
```

MattToast (approved):
A couple of small change requests but nothing directly related to this PR and not worth holding up approval over. Feel free to handle them here if you'd like or kick them out into new tickets!

LGTM!

src/python/module/smartredis/client.py
billschereriii: Done

src/python/module/smartredis/client.py
billschereriii: Done

src/python/module/smartredis/client.py
billschereriii: Done

PR #421, opened 2023-10-26
Add cluster to KW check in client constructor

billschereriii:
This lets a call like `c = Client(cluster=False)` work as expected


MattToast (approved):
LGTM! Thanks for the quick fix!!

PR #427, opened 2023-11-15
Make Dataset::add_tensor() const correct

billschereriii:
The `data` parameter of Dataset::add_tensor() is now properly marked as const. This required const correctness updates to all the methods that are called as part of effecting add_tensor().


MattToast (approved):
LGTM! Thanks for handling this, the immutability makes the API feel much safer!!

al-rigazzi (approved):
LGTM! Thanks for this super-quick response to a user issue!

PR #434, opened 2023-12-05
Make namespace declarations consistent

billschereriii:
The declaration of the `py` namespace was inconsistent. In pyclient.h, it was declared inside the SmartRedis namespace, but in the other py*.h headers, it was declared outside. In this PR, we make the declaration consistent across all of the py*.h headers.


mellis13 (approved):
Thanks for making this consistent -- LGTM

PR #436, opened 2023-12-14
Unpin Intel compiler

billschereriii:
Apply a proper fix to enable use of the Intel OneAPI compiler.


billschereriii:
The underlying issue here is that the Intel Fortran compiler, ifx, is on release 2024.0, but the C/C++ compilers are on release 2023.3. Environment configuration for the compilers is done via `/opt/intel/oneapi/setvars.sh`. The problem is, the version of that script that sets up the right paths for the Fortran compiler is overwritten by the version that sets up the right paths for the C/C++ compilers when we install them. So we have to invoke `setvars.sh` TWICE: once after installing the Fortran compiler, and then once again after installing the C/C++ compilers. This gets all the right paths into play. Invoking it a second time generates an error unless you add the `--force` command-line paramater.

al-rigazzi (approved):
LGTM, very strange behavior indeed!

ashao (approved):
Looks great thanks for tracking this down!

PR #443, opened 2023-12-18
Utilize mutable fields to make Dataset getter methods const

billschereriii:
The following methods have been updated:

- get_tensor()
- get_meta_scalars()
- get_meta_strings()
- get_tensor_names()

Additionally, a few uses of `std::unordered_map::operator[]` have been replaced with calls to `std::unordered_map::at()` since the latter is const but the former is not. This also required adding an exception handler since `std::unordered_map::at()` throws an exception rather than returning null if the requested element is not found

### DIFF
--- a/src/cpp/metadata.cpp
+++ b/src/cpp/metadata.cpp
@@ -270,7 +270,7 @@ std::vector<std::string> MetaData::get_field_names(bool skip_internal) const
 void MetaData::get_field_names(char**& data,
                                size_t& n_strings,
                                size_t*& lengths,
-                               bool skip_internal /*= false*/)
+                               bool skip_internal /*= false*/) const
 {
     // Retrieve the names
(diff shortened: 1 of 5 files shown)


src/cpp/metadata.cpp
```diff
@@ -475,11 +478,14 @@ void MetaData::_get_numeric_field_values(
...
-    SharedMemoryList<T>& mem_list)
+    SharedMemoryList<T>& mem_list) const
 {
     // Make sure the field exists
-    MetadataField* mdf = _field_map[name];
-    if (mdf == NULL) {
+    MetadataField* mdf = NULL;
+    try {
+        mdf = _field_map.at(name);
+    }
```
MattToast, line 487: Good catch! thanks for this fix btw!!

MattToast (approved):
LGTM! Thanks for making the API easier for end users to work with!!

PR #444, opened 2023-12-19
Resolve linting issue

billschereriii:
Avoid wildcard import by reformatting import. Avoid apparent non-use of errors module  by restricting search of pybind error name to errors module.

### DIFF
--- a/src/python/module/smartredis/util.py
+++ b/src/python/module/smartredis/util.py
@@ -29,7 +29,6 @@ from functools import wraps
 
 import numpy as np
 from . import error
-#from .error import *  # pylint: disable=wildcard-import,unused-wildcard-import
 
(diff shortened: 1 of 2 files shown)


src/python/module/smartredis/util.py
```diff
@@ -28,7 +28,8 @@
 from functools import wraps
 
 import numpy as np
-from .error import *  # pylint: disable=wildcard-import,unused-wildcard-import
+from . import error
+#from .error import *  # pylint: disable=wildcard-import,unused-wildcard-import
```
MattToast, line 32: Should be able to drop this dead code as well!

MattToast (approved):
One last small ask, but nothing worth holding up approval over. Otherwise LGTM!!

PR #448, opened 2024-01-09
Improvements to Compiler Compatibility and CI/CD Improvements

ashao:
This PR includes a number of changes that expands the testing coverage and enables SmartRedis to be built with the Nvidia HPC compiler suite.

- The Nvidia HPC toolchain was not quite compiling correctly because some of our dependencies implicitly assume that GCC is being used to compile. This was causing spurious errors do to the fact that these warnings that were interpreted as errors within the dependencies' build steps. This was solved by enforcing that all dependencies are compiled using the system's GCC (and can be overwritten by the user by setting DEPS_CC and DEPS_CXX). Additionally this adds the Nvidia compiler chain to the Github Actions CI/CD.
- Bumping the GCC versions also revealed some innocuous but standard-violating parts of the SmartRedis codebase (e.g. assigning negative values to unsigned integers, unreachable codeblocks)
- The version of Redis is now consistent with the version we use with SmartSim
- The container version was bumped to use Ubuntu 22.04 in order to start testing GCC 12
- The installation of the Intel compilers (and other third party dependencies) via apt was updated to conform with each provider's current instructions
- The version of Torch for testing is now pinned to 1.11.0, the version that SmartSim currently supports


.github/workflows/run_tests.yml
```diff
@@ -84,39 +86,58 @@ jobs:
...
-          echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list &&
-          sudo apt-get update -y &&
-          sudo apt-get install -y intel-oneapi-compiler-fortran &&
+          wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB \
+            | gpg --dearmor | sudo tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null &&
+          echo "deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main" \
+            | sudo tee /etc/apt/sources.list.d/oneAPI.list &&
+          sudo apt update -y && \
+          sudo apt install -y intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp \
+            intel-oneapi-mpi intel-oneapi-mpi-devel intel-oneapi-mkl &&
```
billschereriii, line 116: This is the preferred way to install the intel toolchain in CI/CD now?

.github/workflows/run_tests.yml
```diff
@@ -58,13 +58,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04] # cannot test on macOS as docker isn't supported on Mac
+        os: [ubuntu-22.04] # cannot test on macOS as docker isn't supported on Mac
         rai_v: [1.2.7] # versions of RedisAI
-        py_v: ['3.7.x', '3.8.x', '3.9.x', '3.10.x'] # versions of Python
-        compiler: [intel, 8, 9, 10, 11] # intel compiler, and versions of GNU compiler
+        py_v: ['3.8.x', '3.9.x', '3.10.x'] # versions of Python
+        compiler: [nvhpc-23-11, intel, gcc-11, gcc-12] # intel compiler, and versions of GNU compiler
```
billschereriii, line 64: We're identifying other vendors' toolchains by vendor name, not product name. Might make subsequent code more readable if it were to say "nvidia"

.github/workflows/run_tests.yml
```diff
@@ -58,13 +58,15 @@ jobs:
...
+        link_type: [Static, Shared]
     env:
-      FC: gfortran-${{ matrix.compiler }}
-      GCC_V: ${{ matrix.compiler }} # used when the compiler is gcc/gfortran
+      COMPILER: ${{ matrix.compiler }} # used when the compiler is gcc/gfortran
```
billschereriii, line 67: Are we going to need to differentiate like this for NVIDIA as well?

.github/workflows/run_tests.yml
```diff
@@ -146,52 +168,23 @@ jobs:
...
-                $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null &&
-          # Install and update
-          sudo apt-get -y update && \
-          sudo apt-get -y install iputils-ping docker-ce docker-ce-cli containerd.io redis-tools=6:6.2.5-1rl1~focal1 redis-server=6:6.2.5-1rl1~focal1 &&
-          CONTAINER_NAME="redisai_$RANDOM" &&
-          sudo docker create --name $CONTAINER_NAME --rm redislabs/redisai:${{ matrix.rai_v }}-cpu-bionic && \
-          sudo mkdir -p /usr/lib/redis/modules/ &&
-          sudo docker cp $CONTAINER_NAME:/usr/lib/redis/modules/redisai.so /usr/lib/redis/modules &&
-          sudo docker cp $CONTAINER_NAME:/usr/lib/redis/modules/backends/ /usr/lib/redis/modules/ &&
+          make test-deps &&
```
billschereriii, line 172: I didn't do this before because it is slower than grabbing from the docker container. As it is, explicitly calling make on test-deps is redundant since it will occur as needed with any `make test*` step

.github/workflows/run_tests.yml
```diff
@@ -146,52 +168,23 @@ jobs:
...
-      - name: Install MPI for parallel examples -- for GCC, we can just use the std install
-        if: contains(matrix.os, 'ubuntu') && !contains(matrix.compiler, 'intel')
-        run: sudo apt-get install -y mpich
-      - name: Test static build -- GNU compilers
-        if: contains(matrix.os, 'ubuntu') && !contains(matrix.compiler, 'intel')
-        run: make build-examples SR_LINK=Static SR_FORTRAN=ON
-      - name: Test static build -- intel compilers
-        if: contains(matrix.os, 'ubuntu') && contains(matrix.compiler, 'intel')
-        run: make build-example-serial SR_LINK=Static SR_FORTRAN=ON
+          echo "$PWD/third-party/redis/src" >> $GITHUB_PATH
```
billschereriii, line 174: Why is this needed? The test drivers know how to find redis utils from the SR root

.github/workflows/run_tests.yml
```diff
@@ -146,52 +168,23 @@ jobs:
...
+
+      # Make sure the examples work
+      - name: Run examples
+        run: |
+          make test-examples SR_FORTRAN=ON SR_PYTHON=ON SR_LINK=$LINK_TYPE \
+          SR_TEST_PORT=7000 SR_TEST_REDISAI_VER=v${{ matrix.rai_v }}
```
billschereriii, line 180: I didn't do this before so that we could have the examples and static builds run in parallel with the main tests. Adding them here extends the time for each runner. (I didn't look to see how long the runs are taking now though, so maybe it's not so bad.)

I know we were only running the examples with some compiler chains previously. This may be the source of some of your failures

.github/workflows/run_tests.yml
```diff
@@ -58,13 +58,15 @@ jobs:
...
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04] # cannot test on macOS as docker isn't supported on Mac
+        os: [ubuntu-22.04] # cannot test on macOS as docker isn't supported on Mac
         rai_v: [1.2.7] # versions of RedisAI
-        py_v: ['3.7.x', '3.8.x', '3.9.x', '3.10.x'] # versions of Python
-        compiler: [intel, 8, 9, 10, 11] # intel compiler, and versions of GNU compiler
+        py_v: ['3.8.x', '3.9.x', '3.10.x'] # versions of Python
+        compiler: [nvhpc-23-11, intel, gcc-11, gcc-12] # intel compiler, and versions of GNU compiler
+        link_type: [Static, Shared]
```
billschereriii, line 65: This doubles the number of runners needed, which in turn will make the CI/CD process take a much longer amount of time to complete since it will be pushing us beyond the available amount

Makefile
```diff
@@ -85,7 +87,7 @@ help:
 # help:
 # help: SR_BUILD {Release, Debug, Coverage} -- optimization level for the build
 # help: SR_LINK {Shared, Static} -- linkage for the SmartRedis library
-# help: SR_PEDANTIC {OFF, ON} -- GNU only; enable pickiest compiler settings
+# help: SR_PEDANTIC {OFF, ON} -- GNU only; enable pickiest compiler settings, currently fails do to warnings
```
billschereriii, line 90: do => due

Makefile
```diff
@@ -216,7 +218,7 @@ build-example-serial: lib
 build-example-parallel: lib
 	@cmake -S examples/parallel -B build/$(SR_BUILD)/examples/$(SR_LINK)/parallel \
 		-DSR_BUILD=$(SR_BUILD) -DSR_LINK=$(SR_LINK) -DSR_FORTRAN=$(SR_FORTRAN)
-	@cmake --build build/$(SR_BUILD)/examples/$(SR_LINK)/parellel
+	@cmake --build build/$(SR_BUILD)/examples/$(SR_LINK)/parallel
```
billschereriii, line 221: Yikes!

Makefile
```diff
@@ -44,6 +44,8 @@ CATCH2_URL := https://github.com/catchorg/Catch2.git
 CATCH2_VER := v2.13.6
 LCOV_URL := https://github.com/linux-test-project/lcov.git
 LCOV_VER := v2.0
+DEP_CC := gcc
+DEP_CXX := g++
```
billschereriii, line 48: If these are used only for third party dependencies, add a comment here to explain that

setup.cfg
```diff
@@ -44,7 +44,7 @@ dev =
     black==23.3.0
     isort==5.6.4
     pylint>=2.10.0
-    torch>=1.7.1
+    torch==1.11.0
```
billschereriii, line 47: This feels unrelated to the rest of the changes here

src/cpp/client.cpp
```diff
@@ -1226,7 +1226,7 @@ void Client::set_data_source(std::string source_id)
     // Validate the source prefix
     bool valid_prefix = false;
     size_t num_prefix = _get_key_prefixes.size();
-    size_t save_index = -1;
+    size_t save_index = 0;
```
billschereriii, line 1229: If this is a bug it should be in a separate PR

src/cpp/dbnode.cpp
```diff
@@ -32,7 +32,7 @@ using namespace SmartRedis;
 
 // DBNode constructor
 DBNode::DBNode()
-    : address(), name(""), lower_hash_slot(-1), upper_hash_slot(-1)
+    : address(), name(""), lower_hash_slot(0), upper_hash_slot(0)
```
billschereriii, line 35: This should be a separate PR (along with the previous?)

src/cpp/metadata.cpp
```diff
@@ -230,11 +230,9 @@ void MetaData::get_scalar_values(const std::string& name,
         case SRMetadataTypeString:
             throw SRRuntimeException("MetaData.get_scalar_values() "\
                                      "requested invalid MetaDataType.");
-            break;
```
billschereriii, line 233: I'm assuming that these changes are to remove compiler warnings for unreachable code. Should they all be in a separate PR?

utils/launch_redis.py
```diff
@@ -37,15 +37,10 @@ def check_availability(n_nodes, port, udsport):
...
-        rediscli = 'redis-cli'
-    else:
-        rediscli = (
-            pathlib.Path(__file__).parent.parent
-            / "third-party/redis/src/redis-cli"
-        ).resolve()
+    rediscli = (
+        pathlib.Path(__file__).parent.parent
+        / "third-party/redis/src/redis-cli"
+    ).resolve()
```
billschereriii, line 43: So again, this is removing a performance optimization that we intentionally left in. While I don't object, it does seem like it should be a separate PR

utils/launch_redis.py
```diff
@@ -70,17 +65,11 @@ def stop_db(n_nodes, port, udsport):
...
```
billschereriii, line 72: Same as previous

utils/launch_redis.py
```diff
@@ -159,43 +148,30 @@ def create_db(n_nodes, port, device, rai_ver, udsport):
...
-        redisserver = (
-            pathlib.Path(__file__).parent.parent
-            / "third-party/redis/src/redis-server"
-        ).resolve()
-    rediscli = "redis-cli" if is_cicd else os.path.dirname(redisserver) + "/redis-cli"
+    redisserver = (
+        pathlib.Path(__file__).parent.parent
+        / "third-party/redis/src/redis-server"
+    ).resolve()
+    rediscli = os.path.dirname(redisserver) + "/redis-cli"
```
billschereriii, line 155: and again here

billschereriii (changes requested):
This is at least four PRs rolled together into one, which makes it a bit tough to review:

1. NVIDIA compiler support
2. Compiler warning optimization
3. Intel compiler installation in CI/CD
4. CI/CD changes to eliminate RedisAI installation from Docker container and to move static build and examples into the main "always test"

Changes 1-3 individually look fine, I just think they should have been in separate PRs.
Change 4, however, is probably extending the CI/CD time on commit by over a factor of 2 (though to be fair I haven't actually looked to confirm that this is the case). There probably needs to be a group discussion as to whether that's acceptable in view of expected incoming development.

Additional comments:

1. Not all tests are currently passing. This needs to be fixed before proceeding
2. The change log is not yet updated

.github/workflows/run_tests.yml
ashao: Yes it also results in a pretty minimal installation

Makefile
```diff
@@ -518,10 +522,10 @@ install/lib/libredis++.a: third-party/redis-plus-plus
...
 	mkdir -p compile && \
 	cd compile && \
-	(cmake -DCMAKE_BUILD_TYPE=Release -DREDIS_PLUS_PLUS_BUILD_TEST=OFF \
-		-DREDIS_PLUS_PLUS_BUILD_SHARED=OFF -DCMAKE_PREFIX_PATH="../../../install/lib/" \
-		-DCMAKE_INSTALL_PREFIX="../../../install" -DCMAKE_CXX_STANDARD=17 \
-		-DCMAKE_INSTALL_LIBDIR="lib" -DCMAKE_C_COMPILER=$(DEP_CC) -DCMAKE_CXX_COMPILER=$(DEP_CXX) .. )&& \
+	cmake -DCMAKE_BUILD_TYPE=Release -DREDIS_PLUS_PLUS_BUILD_TEST=OFF \
+		  -DREDIS_PLUS_PLUS_BUILD_SHARED=OFF -DCMAKE_PREFIX_PATH="../../../install/lib/" \
+		  -DCMAKE_INSTALL_PREFIX="../../../install" -DCMAKE_CXX_STANDARD=17 \
+		  -DCMAKE_INSTALL_LIBDIR="lib" -DCMAKE_C_COMPILER=$(DEP_CC) -DCMAKE_CXX_COMPILER=$(DEP_CXX) .. && \
```
billschereriii, line 528: I can't remember why this was done in a subshell previously

.github/workflows/run_tests.yml
ashao: Good point, will follow up with the rest of the group

.github/workflows/run_tests.yml
ashao: The container that we were extracting RedisAI from uses Redis 6, whereas SmartSim is currently using Redis 7. Thus, these modules are now inconsistent with the redis-server that we build. While we could potentially continue testing with Redis 6, upgrading to Redis 7 ensures consistency with SmartSim

.github/workflows/run_tests.yml
ashao: Good point I'll track this down

ashao: This was indeed not needed

.github/workflows/run_tests.yml
ashao: The specification of these versions is now done within the setup section of each compiler chain. The only one that doesn't this to be done is the NVHPC compiler chain because it is installed as a single package (and not individually like Intel and GCC).

ashao:
Thanks @billschereriii for your careful feedback. I have addressed your comments (some of which we discussed offline). One outstanding issue is the reduced breadth of the Github Actions. I did try to expand the matrix as a post-merge action, however previous versions of the Intel compiler set up the compiler environment differently. This requires a non-trivial solution, so will be addressing this in a future PR.

doc/changelog.rst
```diff
@@ -8,16 +8,30 @@ To be released at some future point in time
 
 Description
 
+- Tweak the build system to enable building SmartRedis with Nvidia's toolchain
```
billschereriii, line 11: Nvidia's HPC toolchain?

doc/changelog.rst
```diff
@@ -8,16 +8,30 @@ To be released at some future point in time
...
+- Improvements/upgrades to the container used for Github actions
+- Rework code to avoid compiler warnings
```
billschereriii, line 13: Code updates to resolve warnings?

doc/changelog.rst
```diff
@@ -8,16 +8,30 @@ To be released at some future point in time
...
 - Added developer documentation on how to run a single test case and eliminated duplicative environment variables
 - Resolve a linting issue with pybind-to-python error propagation
 - Use mutable fields to enable Dataset get methods that store memory to be marked const
 
 Detailed Notes
 
+- Change the order of arguments in our MakeFile to ensure that all dependencies are compiled with GCC (PR448_)
+- Add new user-configurable parameters DEP_CC, DEP_CXX to control which compiler is used to build dependencies (PR448_)
+- Bump the container version to Ubuntu 22.04 to be able to start testing GCC 12 (PR448_)
```
billschereriii, line 22: This should say CI/CD in it somewhere

doc/changelog.rst
```diff
@@ -8,16 +8,30 @@ To be released at some future point in time
...
+- Bump the versions of GCC used to only the currently maintained versions (PR448_)
```
billschereriii, line 23: Restrict? And this should say CI/CD in it to keep a customer from thinking we're not letting them use gcc8 any more

doc/changelog.rst
```diff
@@ -8,16 +8,30 @@ To be released at some future point in time
...
+- Add NVHPC to the testing matrix (PR448_)
+- Test the shared/static compilations and examples with all compilers (PR448_)
+- Compile Redis and RedisAI and use those versions in testing instead of extracting from a container (PR448_)
+- Bump the version of Redis used in testing to 7.0.5, the same version as we use with SmartSim (PR448_)
+- Pin the Torch version to 1.11.0, the same as supported in SmartSim (PR448_)
+- Ameliorate some compiler warnings related that were flagged in GCC 12 (unreachable code blocks, invalid values for the type) (PR448_)
```
billschereriii, line 29: suggest replacing "invlaid values for the type" with "signed/unsigned mismatches"

doc/install/lib.rst
```diff
@@ -25,6 +25,24 @@ installed in ``SmartRedis/install/include/``.  The library installation can be
...
 
+The above command will build SmartRedis with the system's default compilers. To
+specify a specific compler users can set the ``CC``, ``CXX``, and ``FC`` variables.
+
+.. code-block:: bash
+
+    # GCC Toolchain
+    make lib-with-fortran CC=gcc CXX=g++ FC=gfortran
+
+    # Nvidia Toolchain
```
billschereriii, line 36: Nvidia (HPC) Toolchain

doc/installation.rst
```diff
@@ -33,6 +33,7 @@ SmartRedis is tested with the following compilers on a daily basis:
     * - Compilers (tested daily)
     * - GNU (GCC/GFortran)
     * - Intel oneAPI (icx/icpx/ifort)
+    * - NVHPC
     * - Apple Clang
```
billschereriii, line 37: Might need to move Clang to the less frequent section now

billschereriii (approved):
Several opportunities for textual clarification, but I have full confidence in your ability to review these appropriately.

Thanks for taking on these changes. It will be a definite improvement to the library to have them in!

I do still recommend conferring with the group on the topic of which compilers to include or not include in CI/CD testing.

doc/changelog.rst
ashao: Agreed: Prefixed these with CI/CD (since most people's eyes will just glaze over)

doc/installation.rst
ashao: I don't want to remove this yet, since it still does compile. Plus, I think because the container isn't used anymore, we might be able to restore the Apple Clang compiler chain.

doc/install/lib.rst
ashao: Agreed, I did put in a table instead

PR #450, opened 2024-01-19
Add Python 3.11 and remove Python 3.7

ashao:
Python 3.7 has reached end of life. Additionally, work has been done on the SmartSim side to confirm that updated versions of the ML backends which do support 3.11 are compatible with SmartSim and SmartRedis. This PR thus removes support for 3.7 and adds support for 3.11

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -60,7 +60,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04] # cannot test on macOS as docker isn't supported on Mac
         rai_v: [1.2.7] # versions of RedisAI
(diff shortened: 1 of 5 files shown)


MattToast (changes requested):
Looks great! Just a couple of minor requests to tag on:
- Can you update the `.github/workflows/run_tests.yml` to include Py 3.11?
- Can you update the `doc/overview.rst` to include `3.11` @ ~line 53
- Can you add something to the changelog, just so that we do not forget to call this out when we release, lol?

ashao:
Thanks for the quick review. Let me know if there's anything else that you can see

MattToast (approved):
LGTM pending that one last test!! Thanks for handling this!!

PR #452, opened 2024-02-01
Add 3.11 and drop 3.7 from README python versions

MattToast:
Make the python versions listed in the SmartRedis `README.md` consistent with the `setup.cfg`


ashao (approved):
Good find

PR #453, opened 2024-02-08
Fix bug when sending an array view

ashao:
NumPy tries to return views of arrays for many types of operations. This however can cause difficulties when trying to send subsets or transposes of the array because the underlying buffer that we access via PyBind11 still points to the original array. Thus the result that gets sent to the database itself is the result of striding through the original array, not the individual piece. To fix this, we detect whether the tensor the user is sending is a view and if so, make an explicit copy.


MattToast (approved):
LGTM!!

PR #455, opened 2024-02-09
Increase build space for Github Actions

ashao:
Our Github Actions require us to install Intel and Nvidia compilers which can be quite large. This leads to failures if we completely run out of space on the Github runner. This uses a new Github Action that removes unneeded packages.


ankona (approved):
LGTM!

PR #456, opened 2024-02-12
Add basic concurrency groups

al-rigazzi:
This PR adds concurrency groups to the `run_tests.yml` workflow, to prevent too many runners from being started by the same PR.


MattToast (approved):
LGTM!! Thanks for freeing up our runners!

PR #462, opened 2024-02-15
Remove a vestigial requirements.txt file

ashao:
We are keeping a requirements.txt file around in the SmartRedis repo which has no apparent use. This removes the file to avoid any misunderstandings about where our dependencies come from.


MattToast (approved):
LGTM!! Thanks for keeping the repo clean!

PR #464, opened 2024-02-16
Remove extra put tensor call in python client

ashao:
An accidental put_tensor call was added when attempting to fix another bug. This is a significant performance concern which is eliminated by removing this line.


MattToast (approved):
Whoops 🤦 LGTM, thanks for the quick turn around!

PR #468, opened 2024-02-16
Fix widths field for list-table

al-rigazzi:
Fixes wrong `widths` value which was preventing table to be renedered.


MattToast (approved):
LGTM!

PR #471, opened 2024-03-12
Fix inconsistency with function parameter variable names

mellis13:
As noted by @al-rigazzi, there is a inconsistency in function argument names for the C-interface of ``ConfigOptions``.  This PR updates the ``is_configured`` function parameter names to be consistent with other functions and the doc strings.  The change in names will not cause a break in user code, so this has not been labeled a breaking change.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -8,14 +8,13 @@ To be released at some future point in time
 
 Description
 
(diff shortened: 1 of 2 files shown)


al-rigazzi (approved):
LGTM! Thanks for this fix!

PR #474, opened 2024-04-01
Add socket time out environment variable

mellis13:
This PR adds the socket timeout parameter as a user-configurable option via environment variables.  This is important because for commands such as ``set_model`` with large models, the internal socket timeout may be exceeded and multiple retries will have no positive affect because more time is needed to complete the action.


ashao (approved):
Looks good to me. I'm going to put a followup ticket to add the list of SmartRedis configurable parameters to the documentation. They're a bit hard to find at this point and should be surfaced.

PR #475, opened 2024-04-02
Update CI for Intel

ashao:
The Github Actions CI was failing because the Intel OneAPI C-compiler, icx, could not be found. This issue was tracked down to the fact that if the versions are not consistent between all the packages, the Intel-provided script that sets up the environment misses some components. This updates the CI to install the entirety of the base and HPC kits for Intel OneAPI. While this takes a significant amount of storage, this is "safest" option for now.


mellis13 (approved):
Thank you for the fix!

PR #476, opened 2024-04-05
Fix minor warnings in C++ client

ashao:
Warnings were being through compiling the library with some of the pickiest compiler flags. The defects that were fixed:
- C++ headers/sources now end with a newline (required by the C++98 standard, deprecated in C++17). Despite complying with the SmartRedis standard, this was fixed to ensure that applications < C++17 that need to include the header files do not throw the errors
- A virtual deconstructor was added for `SRAddress`
- Incorrect inclusion of `stdlib.h`

Additionally, the `SR_PEDANTIC` flag was re-enabled to ensure that additional defects do not creep into the codebase.


include/address.h
```diff
@@ -52,6 +52,11 @@ class SRAddress
     */
     SRAddress() : _is_tcp(true), _tcp_host(""), _tcp_port(0) {}
 
+    /*!
+    *   \brief SRAddress default destructor
+    */
+    virtual ~SRAddress() {}
```
mellis13, line 58: I think this might be the wrong solution.  This would be the ``SRAddress`` destructor doesn't do anything.   I think the error you were seeing is misleading and has nothing to do with the ``SRAddress`` class.  There is no dynamically allocated memory inside of this class that would require a custom destructor.

I think the real problem is in ``sharedmemorylist.tcc``.  I think the problem is that we allocate memory for a shared pointer using ``new[]`` but the shared pointer by default is using the ``delete`` operator to clean up memory when it should use ``delete[]``.  If we put a function like 

```c++
// Define memory deletion operator (delete[]) to
// match the memory allocation operator (new[])
class PointerDeletion {
    public:
        template<class T>
        void operator()(T* ptr) const {
            delete[] ptr;
        }
};
```

and then change line 36 of that file (https://github.com/CrayLabs/SmartRedis/blob/51c3dc6527df9ee9895678b4c534f9fe2b2c763c/include/sharedmemorylist.tcc#L36) to be:

```c++
    std::shared_ptr<T> s_ptr(ptr, PointerDeletion());
```
The warning will go away.  Do you think that is correct?

mellis13 (changes requested):
Thanks for chasing down these warnings!  One small correction to discuss but otherwise good.

include/address.h
ashao: I followed up on your suggestion. Inspecting the code, I tend to agree with you that you've identified a problem (albeit a pedantic one, but a real one nonetheless). I'll run the full suite tomorrow to confirm that the change holds

mellis13 (approved):
(pending tests passing) LGTM!

PR #477, opened 2024-04-07
Remove unused parameter model name in _get_model_script_db()

mellis13:
This PR removes an unused parameter in ``RedisCluster::_get_model_script_db()``.  The comments for the function were also corrected to reflect the actual behavior of the function.  It is worth noting that _get_model_script_db is not called currently in SmartRedis, but it was not removed because it was a precursor implementation for optimizing model execution.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -8,16 +8,20 @@ To be released at some future point in time
 
 Description
 
+- Remove unused parameter in internal redis cluster method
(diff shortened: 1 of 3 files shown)


ashao (approved):
Looks good. Thanks for taking care of this

ashao (approved):
Looks good.  Thanks for clarifying the docstring too.

PR #478, opened 2024-04-07
Enforce matching type for dataset::unpack_tensor

mellis13:
``Client.unpack_tensor()`` enforces that the user-provided ``TensorType`` matches the internally known type.  This check was not added to ``DataSet::unpack_tensor()``, although the information is available.  This check was added to resolve a compiler warning regarding the unused ``TensorType`` function parameter.

### DIFF
--- a/src/cpp/dataset.cpp
+++ b/src/cpp/dataset.cpp
@@ -153,6 +153,7 @@ void DataSet::unpack_tensor(const std::string& name,
     LOG_API_FUNCTION();
 
     _enforce_tensor_exists(name);
+    _enforce_tensor_type(name, type);
     _tensorpack.get_tensor(name)->fill_mem_space(data, dims, mem_layout);
 }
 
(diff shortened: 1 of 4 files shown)


src/cpp/dataset.cpp
```diff
@@ -378,6 +379,22 @@ inline void DataSet::_enforce_tensor_exists(const std::string& tensorname) const
     }
 }
 
+// Check that the provided tensor type matches the internal tensor type
+inline void DataSet::_enforce_tensor_type(const std::string& tensorname,
+                                          const SRTensorType& type) const
+{
+    _enforce_tensor_exists(tensorname);
+
+    SRTensorType known_type = _tensorpack.get_tensor(tensorname)->type();
```
ashao, line 388: I'm a little rusty on how `get_tensor` is implemented, so wanted to double check. IIRC it just returns a pointer to the existing tensor, so the extra additional call to `get_tensor` has a negligible impact on performance, correct?

ashao (approved):
One thing to double-check which comes out of my lack of knowledge, but it otherwise looks good. Thanks for taking care of this

src/cpp/dataset.cpp
mellis13: Yes, that is correct.  The underlying data structure is ``std::unordered_map<std::string, TensorBase*>`` and we are just doing an ``at`` call on that unordered_map (getting a pointer).

PR #479, opened 2024-04-07
Remove unused memory layout constructor parameter for TensorBase.

mellis13:
The ``TensorBase`` object has an ``SRMemoryLayout`` parameter that is not used.  The ``SRMemoryLayout`` is only used by the child ``Tensor<T>`` class.  Because all tensors are stored internally has contiguous memory, there is no need to have it stored as a member variable.  It has been removed to avoid compiler warnings.


ashao (approved):
Thanks for taking care of this, looks good to me.

PR #482, opened 2024-04-09
[no-merge] Drop python 3.8

AlyssaCote:
Python 3.8 is no longer going to be supported because it's near end of life. I've gone through and removed/updated everything that I could find! I added no-merge because I am not sure if there's a specific order that we need to merge all the repos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ SmartRedis provides clients in the following languages:
 
 | Language   | Version/Standard                               |
 |------------|:----------------------------------------------:|
-| Python     |   3.8, 3.9, 3.10, 3.11                         |
+| Python     |   3.9, 3.10, 3.11                         |
 | C++        |   C++17                                        |
 | C          |   C99                                          |
 | Fortran    |   Fortran 2018 (GNU/Intel), 2003 (PGI/Nvidia)  |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 88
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -7,7 +7,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 88
-target-version = ['py37', 'py38', 'py39', 'py310']
+target-version = ['py39', 'py310']
```
mellis13, line 10: @MattToast @ashao @AlyssaCote Is there a reason we don't have py311 here?

MattToast: I'd vote to sneak it in on this PR since it shouldn't hurt anything

AlyssaCote: i'm on it!

README.md
```diff
@@ -34,7 +34,7 @@ SmartRedis provides clients in the following languages:
 
 | Language   | Version/Standard                               |
 |------------|:----------------------------------------------:|
-| Python     |   3.8, 3.9, 3.10, 3.11                         |
+| Python     |   3.9, 3.10, 3.11                         |
 | C++        |   C++17                                        |
```
MattToast, line 38: Very pedantic style thing, but my OCD will not forgive me if I do not call it out: Can we realign these pipes, lol?
```suggestion
| Python     |   3.9, 3.10, 3.11                              |
| C++        |   C++17                                        |
```

AlyssaCote: definitely!

ashao (approved):
No comments other than the ones that @MattToast and @mellis13 picked up on. Only other request is to put up a ticket explaining where bumping to Ubuntu 22.04 in the other actions failed

PR #489, opened 2024-04-30
Readthedocs styling error from headers

amandarichardsonn:
Correct styling issues that was showing up in readthedocs.


AlyssaCote (approved):
Changes look good! Can you change the title to be "Fix changelog header styling", or something like that? Other than that, LGTM!

PR #491, opened 2024-05-14
Fix build wheel

mellis13:
Compiler errors have shown up in the build wheel for SmartRedis.  This PR aims to fix those compiler errors.


ankona (approved):
Great catch!

PR #497, opened 2024-06-17
Build dependencies as external projects

ashao:
Previously we were using a Makefile to download and install third-party dependencies for the library compilation (hiredis, redis++, and PyBind). This PR migrates this functionality to the `CMakeLists.txt` by adding hiredis and redis++ as `ExternalProject`s and by using cmake to download the pybind headers.


mellis13 (approved):
Thank you for making SmartSim compatible as an external project!

PR #505, opened 2024-09-12
Add gcc-14 support

mellis13:
This PR fixes a compilation error that will occur with gcc-14.

Note that gcc-14 was not added to the testing infrastructure because that would require the runners being bumped to ubuntu 24.04 which had some domino effects.  This will be addressed in a future PR.


al-rigazzi (approved):
LGTM, thanks for this quick bug fix!

PR #514, opened 2024-09-26
Fix memory leak in Fortran

ashao:
The dataset object, if used in a loop, would leave memory dangling. To alleviate this, a `final` procedure has been implemented. Fortran compilers, however, are notoriously bad at detecting when an object goes out of scope and to destroy them automatically. We thus also provide an explicit ``destructor`` type procedure.


MattToast (approved):
LGTM!!

PR #520, opened 2024-10-30
Reenable move semantics and suppress warnings

mellis13:
This PR makes two small changes to the SmartRedis code:

1. The ``std::random_device`` instance is not stored as a member variable on the ``RedisServer`` class.  The ``std::random_device`` is not movable or copyable, and as a result, it was implicitly disabling move semantics for the ``Client`` and the ``RedisServer`` derivates. 
2. ``Exception`` class methods have been marked with the appropriate ``const override`` keywords to suppress compiler warnings.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,5 +1,20 @@
 # Changelog
 
+### Develop
+
+To be released at a future time.
+
+Description
+
+-  Reenable move semantics and fix compiler warnings.
+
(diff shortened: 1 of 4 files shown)


al-rigazzi (approved):
LGTM, thanks for this fix!

ashao (approved):
Thanks for polishing this

PR #521, opened 2024-10-30
Enable putting bytes into the database via C++ and Python

mellis13:
This PR adds the ability to put raw byte strings into the database.  This PR covers the C++ and Python clients and Fortran and C clients will be addressed in a follow-up PR.


src/cpp/client.cpp
```diff
@@ -345,6 +345,82 @@ void Client::put_tensor(const std::string& name,
...
+    // We could set the internal str reply value to NULL and the redisreply destructor
+    // which calls free on it will likely do a no-op, but this isn't proven.
+    n_bytes = reply.str_len();
+    data = malloc(n_bytes);
+    std::memcpy(data, (void*)(reply.str()), n_bytes);
+}
+
+// Get byte data and fill an already allocated array 
+// memory space that has the specified size.
+void Client::unpack_bytes(const std::string& name,
```
ashao, line 384: Currently this requires that the buffer and the reply from Redis are exactly the same size. This might be a little overly restrictive especially because the size of the byte string may not be known a priori. I'd suggest that we change this:
```
void Client::unpack_bytes(const std::string& name,
                          void* data,
                          const size_t n_buffer_size,
                          size_t* obj_size
)
```

The check internally should then be that the buffer size is `>=` to the message reply

ashao (changes requested):
Nicely done! One suggestion with regard to the `unpack_bytes` method in the C++ client

ashao (approved):
Looks good. Thanks!

PR #522, opened 2024-10-31
Fix lint errors

mellis13:
This PR fixes lint issues so that other large PRs are less difficult to review.


al-rigazzi (approved):
LGTM, isolating this in its own PR was definitely a good idea

PR #525, opened 2025-03-04
Fix bug in Fortran dataset destructor

ashao:
Fixes a bug in the Fortran client where (on compilers that support automatic destruction) the program would segfault if the user had manually destroyed the dataset. This behaviour was tracked down to an error in the C-client where only the one-level de-referenced pointer was being set to `NULL` but not the actual double-pointer.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -19,9 +19,9 @@ Detailed Notes
 -  For compilers that automatically call the dataset
    destructor when the object goes out of scope, a segfault
    was being triggered if the user was also explictly calling
-   the destructor. This was partially arising because the
-   while we were freeing the underlying data and pointing to
-   null, the pointer to the object itself was never being
+   the destructor. This was partially arising because
+   although we were freeing the underlying data and pointing to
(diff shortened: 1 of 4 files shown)


doc/changelog.md
```diff
@@ -15,8 +16,16 @@ Description
 -  Reenable move semantics and fix compiler warnings.
 
 Detailed Notes
-
-- Update supported Python versions to 3.10, 3.11, and 3.12
+-  For compilers that automatically call the dataset
+   destructor when the object goes out of scope, a segfault
+   was being triggered if the user was also explictly calling
+   the destructor. This was partially arising because the
```
al-rigazzi: `the` at the end of the line should not be there

al-rigazzi (approved):
One minor typo in the changelog, but approving as it looks good otherwise. Thanks for fixing this user issue!

doc/changelog.md
ashao: Danke schoen, grazie mille, et al.

PR #527, opened 2025-06-05
Shift Python support window to 3.10-3.12

al-rigazzi:
Routine update for SmartRedis.
- Add support for Python 3.12
- Update `hiredis` and `redis++` to versions complying with new required `CMAKE_POLICY_VERSION_MINIMUM`
- Update artifact naming to comply with new `upload-artifact@v4` behavior

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -90,5 +90,5 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
+          name: smartredis-wheels-${{ matrix.os }}
           path: ./wheelhouse/*.whl
-          overwrite: true
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -87,7 +87,7 @@ jobs:
         run: brew install cmake || true
 
       - name: Build wheels
-        run: python -m cibuildwheel --output-dir wheelhouse
+        run: CMAKE_POLICY_VERSION_MINIMUM=3.5 python -m cibuildwheel --output-dir wheelhouse
 
       - uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, macos-12, macos-latest]
+        os: [ubuntu-22.04, macos-latest]
         gcc_v: [11] # Version of GFortran we want to use.
     env:
       FC: gfortran-${{ matrix.gcc_v }}
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -51,7 +51,7 @@ jobs:
     env:
       FC: gfortran-${{ matrix.gcc_v }}
       GCC_V: ${{ matrix.gcc_v }}
-      CMAKE_POLICY_VERSION_MINIMUM: 3.5
+      CMAKE_POLICY_VERSION_MINIMUM: "3.5"
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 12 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, macos-12]
+        os: [ubuntu-22.04, macos-latest]
```
MattToast: Not wrong, just curious: is there a reason to have MacOS at latest, but pin Ubuntu to a specific version?

Should this be
```suggestion
        os: [ubuntu-22.04, macos-15]
```
or
```suggestion
        os: [ubuntu-latest, macos-latest]
```
instead?

MattToast: Also, another minor point, probably outside the scope of this ticket since I'm guessing this change is mainly about making an ARM build of SR for Mac rather than overhauling the CI, but iirc `ubuntu-24.04` is now the latest version for linux

.github/workflows/release.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04, macos-12]
+        os: [ubuntu-24.04, macos-14]
```
MattToast: Per my previous comment, these look like good version pins! Can we use these in the `build_wheels.yml` workflow?

CMakeLists.txt
```diff
@@ -106,8 +106,8 @@ set_target_properties(libhiredis PROPERTIES
 # Define redis++ as an external project
 ExternalProject_Add(redis++
     GIT_REPOSITORY https://github.com/sewenew/redis-plus-plus.git
-    GIT_TAG 1.3.10
-    GIT_SHALLOW 1
+    GIT_TAG e30d4c4c557568b01d5adce9df2714de8d3921c2
+    GIT_SHALLOW 0
```
MattToast, line 110: Not wrong, just curious: is there a reason we are grabbing this specific commit over release `1.13.4`? Scrolling through the additional commits from that point I see they updated some of their CMake build, but I didn't see anything that looked too problematic?

examples/examples
MattToast, line 1: Was this simlink intended to be committed? I cannot seem to find a `./smartredis/examples` file/dir from a fresh clone

MattToast (commented):
Looks great!! Just a couple of small questions, mostly for my own curiosity!

Also, depending on how much you want to advertise the python version change before a formal release, you may want to update the python versions listed at:
- `./doc/overview.rst:50`
- `./README.md:37`

Otherwise, seems about ready to go on my end!

CMakeLists.txt
al-rigazzi: The CMake build is problematic. CMake now errors out unless `CMAKE_POLICY_MINIMUM_VERSION` is set to 3.5 or higher. There is no way to circumenvent it in our build pipeline, therefore I grabbed the firtst commit where it was fixed for `redis++`.

examples/examples
al-rigazzi: Ah, thanks, that's some byproduct of building the docs.

.github/workflows/release.yml
al-rigazzi: OK, let's do this for SmartSim too, then.

.github/workflows/build_wheels.yml
al-rigazzi: Yeah, the idea was just to test that Mac ARM always works - it is difficult to stay up to date with all the versioning. Nevertheless, I agree that we should go for `ubuntu-24.04` and `macos-15`.

CMakeLists.txt
MattToast: Ahhhh that makes sense (and should also showcase my complete lack of knowledge of CMake lol)

MattToast (approved):
LGTM!! (pending updating version numbers in `doc/overview.rst` and `README.md`)

PR #530, opened 2025-09-10
Add missing header file

ashao:
When including scalarfield.h into an application, MetadataBuffer was never defined. Add an explicit include to the header file to ensure that applications can build.


MattToast (approved):
LGTM!! Thanks for the catch!

PR #533, opened 2026-03-25
Add compatibility with Numpy 2.x

ashao:
The version of pybind11 was rather old and had known compatibility issues with numpy 2.x. This manifested within SmartRedis with Python clients filling tensors sent to or retrieved from the database with just the value of the first element of the tensor.


MattToast (approved):
LGTM, the CI failures just look like 'Out of Space' problems that we can fix later!!

